Issue #1, opened 2018-11-01
change project naming


jtauber:
The main Django project is still called `basic`. That should be changed to `hedera` everywhere.

arthurian:
Just merged PR #7, so this issue should be all set.

PR #7, opened 2018-11-01
Rename django project from basic to hedera

arthurian:
This PR renames the main Django project from `basic` to `hedera` per #1. 

@jtauber


paltman:
LGTM

Issue #2, opened 2018-11-01
minimal brand changing


jtauber:
We should do minimal changes to the branding to be clearly Hedera. Nothing fancy at this stage but just the site name, homepage heading, etc. Possibly tweak the colour-scheme just to be different too.

arthurian:
Closing this as a result of merging #10.

PR #10, opened 2018-11-02
Minimal branding changes

arthurian:
Tweaked the branding slightly to be clearly Hedera (per issue #2). 

- Switched from perseus red to hedera green (the exact shade/color scheme can be tweaked later).
- Updated language to reflect that it's Hedera, not Scaife Basic.
- Added hedera background image to home page. 

Should look like this:

hederabranding


static/src/scss/_home.scss
```diff
@@ -1,6 +1,8 @@
 body.home {
   section.home-hero {
-    background-color: $gray-800;
+    background-image: url("/site_media/static/images/hedera-ivy.png");
```
paltman: This file hasn't been added to source control.  Can you please add it @arthurian?

paltman: oh looks like it's a `jpg`

arthurian: Good catch @paltman. I think this should be fixed now.

PR #12, opened 2018-11-06
initial vocabulary list models, admin, and sample fixture

jtauber:
Completes #8.

The vocabulary list models are simplified from earlier work-in-progress done for DeepVocabulary by James Harrington and myself.

I've included a sample fixture with lemmas and glosses for the top 318 words in the Greek New Testament (providing 80% coverage).


vocab_list/models.py
```diff
@@ -0,0 +1,55 @@
...
+    owner = models.ForeignKey(
+        getattr(settings, "AUTH_USER_MODEL", "auth.User"),
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE
+    )
+    title = models.CharField(max_length=255)
+    description = models.TextField(blank=True)
+
+    created = models.DateTimeField(auto_now_add=True)
```
paltman: There has been talk for a while about deprecating `auto_now_add` and `auto_now`.  I'm +0 for using `default=timezone.now` instead and then having:

```python
def save(self, *args, **kwargs):
    self.updated = timezone.now()
    return super().save(*args, **kwargs)
```

Could put these two fields and the save method on a base model class.

vocab_list/views.py
```diff
@@ -0,0 +1,4 @@
+from django.shortcuts import render
```
paltman: Can we delete this and `tests.py` until they are needed?

vocab_list/models.py
jtauber: Well, it was frowned upon years ago but looking at the Django 2.1 docs, it seems it's back in favour.

James Harrington's original code had abstract base classes but I felt it overcomplicated things at this stage.

vocab_list/views.py
jtauber: yeah, the views will probably come shortly but may as well remove them both for now.

arthurian:
Looks good @jtauber

PR #16, opened 2018-11-30
Code to enable deployment to a remote development environment on AWS

joshuagetega:
This PR introduces the following changes that enable a deployment to a development environment on AWS:

- A ./Dockerfile containing instructions to build the application and bundle it in a docker image
- A ./buildspec.yml file that triggers the aforementioned Dockerfile instructions
- A ./hedera_nginx.conf that configures nginx for the app
- A ./docker-entrypoint.sh file that runs nginx and gunicorn so that the application can be served up

An infrastructure setup of this project is to be found at https://github.com/Hedera-Lang-Learn/hedera-terraform/tree/aws-codepipeline-ecs-IaC. Thus, a quick way to test the changes introduced in this PR would be to run the `terraform [command]` commands in the aforementioned repo. That repo contains a CodeBuild and CodePipeline that will engage all the PR changes in this PR. Expressed elaborately, the terraform code in the aforementioned repo depends on a proper setup of the ./Dockerfile, ./buildspec.yml, ./hedera_nginx.conf, and ./docker-entrypoint.sh files in this repo. Hence, it might be effective to test the terraform code first, before diving head-first in the PR changes in this repo - just for the sake of catching errors quickly, at least. Please let me know if you have any questions.

### DIFF
--- a/basic/settings.py
+++ b/basic/settings.py
@@ -15,7 +15,9 @@ DATABASES = {
 }
 
 ALLOWED_HOSTS = [
-    "localhost",
+    'localhost',
+    '127.0.0.1',
+    'atg-dev-hedera-lb-2073615562.us-east-1.elb.amazonaws.com'
 ]
 
 # Local time zone for this installation. Choices can be found here:
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ site_media/
 static/dist/
 .DS_Store
 .vscode
+.idea/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM circleci/python:3.7-rc-node
+USER root
+RUN mkdir -p /srv/code
+WORKDIR /srv/code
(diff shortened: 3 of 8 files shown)


basic/settings.py
```diff
@@ -14,9 +14,7 @@
     }
 }
 
-ALLOWED_HOSTS = [
-    "localhost",
-]
+ALLOWED_HOSTS = ['*']
```
arthurian: Setting this to a wildcard disables [security protection on the Host header](https://docs.djangoproject.com/en/2.1/ref/settings/#allowed-hosts). 

It would probably be better to make this an environment variable that supplies the list of allowed hosts, defaulting to localhost when none is provided:

```python
ALLOWED_HOSTS = os.environ.get('DJANGO_ALLOWED_HOSTS', 'localhost').split(' ')
```

I'm not sure what that would mean on the ECS side of things as far as setting that environment variable.

PR #31, opened 2019-03-13
Use dj_database_url for db backend config

joshuagetega:
DATABASE_URL is used in place the canonical django database env variables, e.g. DB_NAME, DB_HOST, etc, for specifying a database backend.
See https://github.com/kennethreitz/dj-database-url for reference.

### DIFF
--- a/docker-entrypoint-dev.sh
+++ b/docker-entrypoint-dev.sh
@@ -2,6 +2,7 @@
 
 export AWS_SECRET_ACCESS_KEY=`aws ssm get-parameter --name /hedera/dev/aws_secret_access_key --with-decryption --output text --query Parameter.Value --region us-east-1`
 export AWS_ACCESS_KEY_ID=`aws ssm get-parameter --name /hedera/dev/aws_access_key_id --with-decryption --output text --query Parameter.Value --region us-east-1`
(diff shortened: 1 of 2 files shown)


joshuagetega:
@paltman, does my use of DATABASE_URL in settings.py look as expected? I have tested it and it works. (Fyi, we store environment variables securely in AWS Systems Manager parameter store - works well with AWS ECS. Therefore, the DATABASE_URL env var is populated by pulling the appropriate value from the aws parameter store, as you will see in the changes in docker-entrypoint-dev.sh and docker-entrypoint-prod.sh.)

hedera/settings.py
```diff
@@ -13,6 +13,8 @@
     "default": dj_database_url.config(default="postgres://localhost/hedera")
 }
 
+DATABASE_URL = os.environ.get("DATABASE_URL")
```
paltman: @joshuagetega this shouldn't be necessary.  `dj_database_url` above will read in `$DATABASE_URL` from env and use that instead of the `default` that is provided, if it exists.

joshuagetega: I see, @paltman . I've removed the line and tested the update. The db connection is made successfully without the line. Thanks for the feedback. Going ahead with the merge ...

joshuagetega:
@paltman , I've made the change. Thanks. If there's nothing else please approve, and I'll go ahead and merge.

paltman:
LGTM

PR #41, opened 2019-04-03
Background the Lemmatization

paltman:
Longer texts can take considerable time to lemmatize and not only is a bad experience for the user but could end in the request timing out.

This PR puts the lemmatization on a background task and immediately returns the user to the list of texts with an indication of the lemmatization status for the user.

It does require some new things on the operations side:

* Redis
* setting `RQ_ASYNC=1` in the web and worker environments
* A Worker (`python manage.py rqworker ${RQ_QUEUES:-default}`)


jtauber:
I'd like to check whether the caching of lookups to the database has regressed

jtauber:
I guess if there's a lot in a text that couldn't be lemmatised, it will be attempted every time. So that might explain it. Would be good to confirm that morpheus isn't called as many times after the first lemmatisation, though.

PR #121, opened 2019-11-20
Bills branch

bilbe:
I added a "stickiness" to the language choice for the create page in lemmatized texts. It now automatically selects the same language as that of the last text submitted by the user . You can, of course, change that. It could be fancier, but it is a nice convenience for everybody not working in... uhm, Greek. 
--Bill B.

### DIFF
--- a/lemmatized_text/views.py
+++ b/lemmatized_text/views.py
@@ -1,3 +1,5 @@
+import logging
+
 from django.db.models import Q
 from django.shortcuts import get_object_or_404, redirect, render
 from django.views.decorators.http import require_POST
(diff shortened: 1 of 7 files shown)


lemmatized_text/views.py
```diff
@@ -31,9 +41,15 @@ def create(request):
...
         lt.lemmatize()
         return redirect("lemmatized_texts_list")
+
+    # the GET is for initial page or getting the cloned from object
     if request.GET.get("cloned_from"):
         cloned_from = get_object_or_404(models.LemmatizedText, pk=request.GET.get("cloned_from"))
-    return render(request, "lemmatized_text/create.html", {"cloned_from": cloned_from})
+
+    # if the text is being cloned then make sure that language is the selected one
+    if cloned_from: select_lang = cloned_from.lang
```
dodget: Excellent.

PR #142, opened 2019-12-16
Fix favicon static

joshuagetega:
This PR highlights 2 changes:

1. Modification of the STATIC_URL and MEDIA_URL. See commit 92988d1.
Previously, the values of both variables were equal to the values of STATIC_ROOT and MEDIA_ROOT respectively. This should not be the case. It makes us manually put static files directly in STATIC_ROOT, which is not great. Instead, we specify a different location in STATIC_URL that we can use to manually put static stuff in. The django `collectstatic` command will get that stuff into STATIC_ROOT automatically.

2. Creation of an app-level 'static' directory and moving the favicons assets from the project-level 'static' directory into the newly-created app-level one. See commit 08f795c. (By 'app' I mean a django app within a django project, e.g. the hedera django app within the hedera django project). This is because the settings.py file is in the hedera django app, which means that the static location specified in STATIC_URL in the very settings.py file is an app-level one.

Given these changes, I have tested and confirmed that the favicon assets get uploaded to the s3 bucket successfully and show up on the website as expected.


IvyJL:
Agreed! Looks good!

Ivy

Issue #138, opened 2019-12-11
Need to be able to delete word lists added in teacher mode


IvyJL:
I uploaded a list, realized it had a mistake, and uploaded a corrected version with the same name. Now my list of lists has two with the same name.

PR #152, opened 2019-12-30
Add the ability to delete vocab lists

paltman:
#138


jtauber:
I think the failure is just an isort issue

PR #166, opened 2020-02-21
Fixed Sentry errors found when uploading vocab lists.

bilbe:
It was just a couple little things but it caused errors when uploading vocab lists.

### DIFF
--- a/vocab_list/models.py
+++ b/vocab_list/models.py
@@ -50,13 +50,20 @@ class VocabularyList(models.Model):
     )
 
     def load_tab_delimited(self, fd):
-        def clean(headword, gloss):
+        def clean(headword, gloss, *args):
+            """Expect a headword and gloss, i.e., two columns,
+               & additional columns will be ingnored."""
             return (headword.strip().strip('"'), gloss.strip().strip('"'))
 
-        lines = [
-            clean(*line.decode("utf-8").strip().split("\t"))
-            for line in fd
-        ]
+        def parse_line(idx, line):
+            try:
+                columns = line.decode("utf-8").strip().split("\t")
+            except UnicodeDecodeError:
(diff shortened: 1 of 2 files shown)


vocab_list/views.py
```diff
@@ -75,6 +75,6 @@ def get_context_data(self, **kwargs):
 
     def form_valid(self, form):
         vl, _ = PersonalVocabularyList.objects.get_or_create(user=self.request.user, lang=form.cleaned_data["lang"])
-        entries = vl.load_tab_delimited(form.cleaned_data["data"], familiarity=form.cleaned_data["rating"])
```
paltman: not sure why this was needed or why it's in the same context as the change to the `VocabularyList` method above as this goes to the `PersonalVocabularyList` which has a `familiarity` keyword argument.

vocab_list/models.py
```diff
@@ -50,13 +50,20 @@ class VocabularyList(models.Model):
     )
 
     def load_tab_delimited(self, fd):
-        def clean(headword, gloss):
+        def clean(headword, gloss, *args):
```
paltman: I guess you hit an issue where because your spreadsheet had more than two columns.  Is that so?

bilbe: That's right. We just wanted to deal with this in a way that didn't cause the process to fail.

vocab_list/views.py
bilbe: Sorry this took so long to get back to you on this. As I remember, we removed the "familiarity=" because `load_tab_delimited` method on the `PersonalVocabularyList` object is not expect a key-value pair, so it was being turned into key-key-value, i.e., one extra level. This was only a problem when uploading a vocabulary list.

PR #167, opened 2020-02-24
GitHub actions

paltman:


paltman:
@bilbe this gets us off CircleCI and uses Github Actions to run CI tasks.  we have found this to be a good bit tighter/neater integration.  The last step before we can merge this though is to go into Settings on this repo, and remove the requirements for circleci to pass since it's now disconnected.  I don't have access to do this but assume you do.

joshuagetega:
@paltman, I deleted the CircleCi webhook from the repository settings. That was the only CircleCi-related setting I came across. In addition, @bilbe turned off building from within the CircleCi web app. That said, I see the "ci/circleci: Build Error" notification has not disappeared from this pull request even after these changes. I'm not sure whether it should, but I think CircleCi checks will not be triggered in future pushes given the aforementioned changes. Thus, feel free to merge the PR.

PR #161, opened 2020-02-04
Weighted percentage of known words

joshuagetega:
This PR introduces a getter method for computing weighted percentage of known words. Also, visually, an extra chart is displayed right below the one that shows raw percentage.

PR #175, opened 2020-04-07
Revert pull request #161 to fix freezing-page problem

joshuagetega:
Pull request #161 was the one introducing a second visualization to display weighted percentage of known words. The reason for undoing the pull request is that the changes introduced seem to be the cause of the browser-page-freeze that happens when you click the vocab list dropdown and select a vocab list. The reason for the freeze is likely that the changes introduced in the pull request are a bit heavy to run on the front-end. I noticed that when you open the Chrome Task Manager and select a vocab list, the CPU usage spikes to 100+ and stays as such indefinitely. Thus, perhaps the computation introduced in the PR is heavy for running on the front-end or there is a bug in the code change in that PR. The real reason is yet TBD. At any rate, reverting the pull request resolves the problem.

Ideally, I would have tracked down the problem in pull request #161 and fixed the bug. But a hotfix is needed for the freezing-page problem. This is the hotfix; i.e. reverting the PR. A proper solution will be provided later.

### DIFF
--- a/static/src/js/app/vuex/getters.js
+++ b/static/src/js/app/vuex/getters.js
@@ -5,37 +5,6 @@ export default {
     const knownTokens = tokens.filter(t => t.inVocabList).length;
     return knownTokens / totalTokens;
   },
-  weightedKnownVocab: (state) => {
-    const tokens = state.tokens.filter(t => t.word === t.word.toLowerCase());
-
-    // Create a set of unique tokens
-    const uniqueTokens = new Set();
-    for (let i = 0; i < tokens.length; i + 1) {
-      if (tokens[i].resolved !== 'na') {
-        if (tokens[i].label) {
-          // Looks like 'label' stores the lemma
-          uniqueTokens.add(tokens[i].label);
-        } else {
-          // If there is no lemma, use the word as is
-          uniqueTokens.add(tokens[i].word);
-        }
-      }
-    }
-
-    // Determine which tokens in the set of unique tokens are known,
-    // and put them in a new set
-    const knownUniqueTokens = new Set();
-    for (let i = 0; i < tokens.length; i + 1) {
-      if (tokens[i].inVocabList) {
-        knownUniqueTokens.add(tokens[i].label);
-      }
-    }
-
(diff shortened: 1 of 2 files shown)


joshuagetega:
@arthurian, @paltman, @jtauber, I am going to merge this right away given it is a hotfix. Just alerting you to inform you of the change.

static/src/js/app/vuex/getters.js
```diff
@@ -5,37 +5,6 @@ export default {
     const knownTokens = tokens.filter(t => t.inVocabList).length;
     return knownTokens / totalTokens;
   },
-  weightedKnownVocab: (state) => {
-    const tokens = state.tokens.filter(t => t.word === t.word.toLowerCase());
-
-    // Create a set of unique tokens
-    const uniqueTokens = new Set();
-    for (let i = 0; i < tokens.length; i + 1) {
```
arthurian, line 13: @joshuagetega Is this the problematic line in question? Looks like there's a typo, since `i + 1` should be `i += 1` or `i++`. Without the assignment, this is just an infinite loop because `i` never increments.

static/src/js/app/vuex/getters.js
```diff
@@ -5,37 +5,6 @@ export default {
...
-          // If there is no lemma, use the word as is
-          uniqueTokens.add(tokens[i].word);
-        }
-      }
-    }
-
-    // Determine which tokens in the set of unique tokens are known,
-    // and put them in a new set
-    const knownUniqueTokens = new Set();
-    for (let i = 0; i < tokens.length; i + 1) {
```
arthurian, line 28: Same comment as above.

static/src/js/app/vuex/getters.js
joshuagetega: Ah, right. That's it. Thanks for catching it. I'm not sure how I did not catch that when I was testing the code after writing it. At any rate, I will rectify this by reintroducing the reverted changes, correcting the loop, testing the changes, and issuing a new PR with the corrected changes.

PR #180, opened 2020-04-22
Batch Lemma Disambiguation

paltman:
### Done

1. automatically mark as resolved when node is selected
2. add highlighting of identical forms

### To Do

3. separate button for "apply all" (once you've disambiguated one)


paltman:
We can certainly merge this and start taking advantage of #1 and #2 before #3 is done (which I'm not sure if I'll get to before our call tomorrow).  @jtauber @bilbe Thoughts?

paltman:
Creating a new issue for 3 so we can merge 1 and 2 into dev now.

PR #187, opened 2020-04-29
Customize font for reading

arthurian:
This PR updates the [Noto](https://www.google.com/get/noto/) font definition so that russian text, particularly accented text, is rendered correctly. 

Example before and after: 

noto-serif

not-serif-russian

(Edited per discussion below)

### DIFF
--- a/hedera/templates/site_base.html
+++ b/hedera/templates/site_base.html
@@ -8,7 +8,7 @@
 
 
 {% block styles %}
-  <link href="https://fonts.googleapis.com/css?family=Noto+Serif:400,400i,700,700i&amp;subset=greek,greek-ext,latin-ext" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


jtauber:
That seems odd given Noto has all the code points. Are you sure we don't just need to change the Google fonts URL to include more subsets?

arthurian:
Oh that's entirely possible. I'll take a look!

jtauber:
I deliberately used Noto because it has such good coverage so I but we're subsetting it too narrowly.

jtauber:
Yeah:
```
<link href="https://fonts.googleapis.com/css?family=Noto+Serif:400,400i,700,700i&amp;subset=greek,greek-ext,latin-ext" rel="stylesheet">
  <link href="https://fonts.googleapis.com/css?family=Noto+Sans:400,700" rel="stylesheet">
```

arthurian:
@jtauber It looks like specifying the subset did that trick just as you suggested. The coverage of the `Noto` font family is quite impressive. Thanks for pointing me in the right direction here. Does the updated commit look good to you?

Issue #137, opened 2019-12-05
words like iuvo, iaceo not lemmatizing


IvyJL:
These words begin with consonantal i, which can also be written j. I thought maybe Morpheus was expecting j, but when I type iaceo with an i into the morpheus web interface (https://www.perseus.tufts.edu/hopper/morph?l=iaceo&la=la#lexicon), it actually returns jaceo with a j. So it seems like Morpheus recognizes both spellings. Not sure where the breakdown is occurring then.

bilbe:
iaceo and jaceo need to resolve to the same lemma. Spelling variations are common and need to be dealt with.

jtauber:
I just confirmed the Morphology API is returning ['jaceo'] and ['juvo'] for iaceo and iuvo so the problem doesn't appear to be morpheus. Will investigate further downstream in our code.

jtauber:
Still in progress, but just to update with the latest findings: the issue seems to be not that the lemmatizer didn't know the word but rather that once the lemma came back, it was not in the lattice (and we changed things to prevent the lemmatizer from creating new nodes except to handle ambiguity).

So the underlying issue is probably that Logeion doesn't have the lemmas in their j- forms as returned by Morpheus but that's the next thing I'm investigation.

jtauber:
On a fresh install, iaceo and jaceo both work, I think because Ivy's lattice links them.
iuvo/juvo both fail because (a) morpheus returns juvo and (b) logeion has iuvo and the combination makes it unrecognised.

We can discuss on the call, but perhaps the best solution is to normalise all j to i for internal matching (which is what Bill was saying all along but I wanted to get to the root cause of the problem).

jtauber:
* let's convert Morpheus's j's coming back to i's.

PR #194, opened 2020-05-12
Normalize i/j and u/v spelling for latin 

arthurian:
This PR is a first pass at normalizing j/i and u/v for internal matching (#137). It just does a simple replacement of j->i, u->v. 

Are there additional rules that need to be considered for this sort of spelling normalization or is this sufficient?


jtauber:
We can discuss on a call but I have questions about how this might interact with the morphology service and also vocabulary lists.

arthurian:
Discussed this PR during today's call and we decided to close this PR since normalization needs more consideration overall.

PR #196, opened 2020-05-18
Bulk disambiguation

paltman:
You can now select "Apply All" after marking a token as resolved if there are other matching tokens in the text to apply the same disambiguation to them in one go.

@jtauber Could you do a review of this for me.  There is a lot of changes to support this that could impact a lot of things.  I have tested it locally and fixed what I've found except for "Add a Lemma".  That's not working but when I went to troubleshoot it I don't see how it could possibly be working currently as the I don't see in the API handler where the data is being handled.  

I'll verify if this is indeed a regression or not later, but wanted to get this PR up and ready for review.  Regression or not, we need to fix the the ability to add a lemma but might not need to hold up this merge.


jtauber:
'Add a Lemma' hasn't worked for a year (if it ever did) so definitely not a regression caused by this.

PR #215, opened 2020-07-21
Fix django ALLOWED_HOSTS for subdomains

joshuagetega:
This PR fixes ALLOWED_HOSTS for subdomains so that dev.hederaproject.org can work without throwing an error.

### DIFF
--- a/hedera/settings.py
+++ b/hedera/settings.py
@@ -37,7 +37,7 @@ ALLOWED_HOSTS = [
     "localhost",
     "127.0.0.1",
     "hederaproject.org",
-    "*.hederaproject.org",
+    ".hederaproject.org",
 ]
 
 ALLOWED_HOSTS += get_ecs_task_ips()


paltman:
Oh wow, I had thought that `*.DOMAIN` worked.  I wonder if it used to work and latest Django changed it or maybe I've been mistaken all this time. :)

joshuagetega:
@paltman, this was my bad. It was ok up until last week when I was working [on this PR](https://github.com/Hedera-Lang-Learn/hedera/pull/212/files), where I noticed the wildcard was missing, and so my regex mind thought the wildcard should be there, so I added it in. Turns out it was right without the wildcard, as per the sentence below in the [django docs](https://docs.djangoproject.com/en/2.2/ref/settings/):

> A value beginning with a period can be used as a subdomain wildcard: '.example.com' will match example.com, www.example.com, and any other subdomain of example.com.

PR #221, opened 2020-07-28
Refactor dockerization for the sake of the LTI version of hedera

joshuagetega:
This PR refactors the dockerization of hedera so that the LTI version can be integrated. It simplifies the dockerization such that both versions (LTI and non-LTI) and both environments (dev and prod) share the same entrypoint. So, instead of 4 entrypoints, we just have 1. This collapsing of entrypoint files is enabled by removing from the entrypoints the awscli commands that were getting environment variables from the parameter store, and injecting said variables into the containers via the task definitions. The variable injection has thus been moved from this source code into the operations code.


web-entrypoint.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
arthurian, line 1: Now that the environment variables are populated by the task definition, this is much more streamlined and generalized.

PR #234, opened 2020-08-12
Add ability to edit/delete entries on vocab list

paltman:
Here is the editing and deleting UI for a vocab list.  Still need to add the ability to add a single entry.

Screen Recording 2020-08-12 at 02 35 46 PM


IvyJL:
Wow! That looks super!

paltman:
Was there anything blocking this from merge?

PR #235, opened 2020-08-17
Lti integration

dodget:
This PR add LTI integration in the following sections:

## Pipfile

- Adding fork of `django-lti-provider`

## settings.py

- Adding `IS_LTI` flag, which should be set in environment variables
- Using `IS_LTI` to prevent sentry from unwittingly collecting student data
- Adding LTI tool configuration settings needed for `django-lti-provider` package

## lti app

- Contributing new `lti` app containing two new views (LtiInitializerView, LtiRegistrationView)
- Authenticating users based on variables provided by an LTI launch if `IS_LTI` is True
- Providing a username registration form/template if the LTI user is new
- Creating a new group or finding the corresponding group at each LTI launch
- Updating the user's group roles at each LTI launch
- Providing a template if a user's session has expired
- Adding utility function to login LTI user
- Providing tests for new views and functionality

## context_processors.py

- Adding `IS_LTI` value to template contexts

## templates

- Using `IS_LTI` to "hide" actions that we don't necessarily want user to take (like change their email address)
- Adding lti registration template, lti failure template

## Notes/Questions

- Because I initially set Hedera up to run as an app in canvas, it required ssl to run locally in order for the app to work
- Now that Hedera launches in a new tab, we probably wont need https to run the app locally, so maybe I can strip it out?
- The use of request.session in `AuthenticationMiddleware` seems fragile. Perhaps there is a better pattern?


dodget:
@arthurian - I'm thinking about your comment on the confusing nature of the LtiInitializerView's `dispatch` method, and the multiple ways it can retrieve the values that it needs -- because you can get there via POST, or redirect. I'm thinking that the logic should be split into the `get` and `post` methods, so then we are more explicit about where we should be finding values for things like email, and course_id (whether it should be in the POST or the session).

hedera/settings.py
```diff
@@ -8,17 +8,20 @@
 from .aws import get_ecs_task_ips
 
 
+IS_LTI = bool(os.environ.get("IS_LTI"))
```
joshuagetega, line 11: Shell environment variable values are read in as strings. Accordingly, the expression `bool(os.environ.get("IS_LTI"))` will always evaluate to `True`, except when the env variable is an empty string, i.e. `IS_LTI=""`. Concretely, even if the env variable `IS_LTI=0` or `IS_LTI=False`, the expression `bool(os.environ.get("IS_LTI"))` will still evaluate to `True`. That would be undesired behavior.

To fix this, one might try:
```
IS_LTI = bool(int(os.environ.get("IS_LTI", "0")))
```
This way, the setting `IS_LTI` will be `True` if and only if the env variable `IS_LTI=1`. Note that a `ValueError` will be thrown if the env var value is not a string that can convert to type `int`. So, a better solution then becomes:
```
try:
    IS_LTI = bool(int(os.environ.get("IS_LTI", "0")))
except ValueError:
    IS_LTI = False
```
Or, succinctly:
`IS_LTI = (os.environ.get("IS_LTI") == "1")`

joshuagetega (approved):
Hi @dodget, this looks good, except for an issue with the `IS_LTI` setting. See my comment on that.

hedera/middleware.py
```diff
@@ -26,6 +30,19 @@ def __call__(self, request):
             if re.match(exemption, request.path):
                 response = self.get_response(request)
                 return response
+        if settings.IS_LTI and not request.user.is_authenticated:
+            try:
+                lti_user = login_existing_user(request)
+            except ObjectDoesNotExist:
+                request.session["lti_email"] = request.POST.get("lis_person_contact_email_primary", False)
```
joshuagetega, line 37: This seems fine to me. For semantics, though, I wonder where the lti param values can be made to default to `None` instead of `False`. For example, `request.session["lti_email"] = request.POST.get("lis_person_contact_email_primary", None)`. Defaulting to `False` might imply that a value of type `bool` is expected. Defaulting to `None` implies that something is expected but none was found.

joshuagetega:
@dodget, regarding your questions:

> Now that Hedera launches in a new tab, we probably wont need https to run the app locally, so maybe I can strip it out?

Perhaps we can leave it in. In the future, it is conceivable we may want the app to open in an iframe as opposed to a new tab. If leaving it in doesn't do any harm, I'm ok with that. That way, we wouldn't need to redo the work of setting up https.

> The use of request.session in AuthenticationMiddleware seems fragile. Perhaps there is a better pattern?

Seems fine to me. I just made a [minor comment](https://github.com/Hedera-Lang-Learn/hedera/pull/235#discussion_r473140417) on that.

hedera/settings.py
dodget: Yeah, I was following the pattern already set in this file for reading env vars. The way that I typically do it is:

`IS_LTI = os.environ.get("IS_LTI", False)` knowing that if I want the value to be false, it has to be set as `IS_LTI=` or removed totally in the environment.

hedera/middleware.py
dodget: Good point

dodget:
@joshuagetega @arthurian 

I appreciate the feedback and ideas. I implemented the majority of your suggestions, and moved the LTI verification/user authentication out of MIDDLEWARE and into the `dispatch` method of the `LtiInitializerView`. I did not add a `LTIRegistrationSuccessView` because at this point, the `LtiInitializerView` does not seem bloated enough to refactor. That just my opinion, however.

Moving the user authentication out of MIDDLEWARE and into the `dispatch` of `LtiInitializerView`, as well as refactoring the `dispatch` method into `get` and `post` has made it more robust and comprehensible overall. The way I am forcing users to `LtiInitializerView` for authentication is by setting `LOGIN_URL = "lti_initializer"` is `IS_LTI` is `True`. It all seems to be working well.

I also removed any need for ssl in local development. I hear you when you say that its work that is already done, but I see it as a potential blocker for a new person beginning local development that isn't totally necessary at this point.

I had to drop a couple of tests, because I need to research how to mock a verified lti launch, but I don't have the time at the moment. Further, the `determine_role` method of `LtiInitializerView` could be improved. Perhaps those two items are two point stories each in an upcoming sprint?

hedera/settings.py
```diff
@@ -8,17 +8,20 @@
...
-    dsn=os.environ.get("SENTRY_DSN"),
-    debug=os.environ.get("SENTRY_DEBUG") == "1",
-    environment=os.environ.get("SENTRY_ENVIRONMENT"),
-    integrations=[DjangoIntegration(), RqIntegration()],
-    # Enables tracing for sentry "Events V2"
-    # https://github.com/getsentry/zeus/blob/764df526f47d9387a03b5afcdf3ec0758ae38ac2/zeus/config.py#L380
-    traces_sample_rate=1.0,
-    traceparent_v2=True,
-)
+if not IS_LTI:
```
paltman, line 14: @dodget @arthurian why is Sentry being disabled here?

hedera/urls.py
```diff
@@ -34,6 +36,10 @@
     path("api/v1/personal_vocab_list/", api.PersonalVocabularyListAPI.as_view()),
     path("api/v1/personal_vocab_list/<int:pk>/", api.PersonalVocabularyListAPI.as_view()),
 
+    path("lti/", include("lti_provider.urls")),
+    path("lti/lti_initializer/", LtiInitializerView.as_view(), name="lti_initializer"),
+    path("lti/lti_registration", LtiRegistrationView.as_view(), name="lti_registration"),
```
paltman, line 41: Can we add the trailing slash for consistency with other urls?

```suggestion
    path("lti/lti_registration/", LtiRegistrationView.as_view(), name="lti_registration"),
```

hedera/settings.py
dodget: There was some concern over exactly what information sentry is collecting from users, and how it may violate student's FERPA rights. We thought it would be the simpler route to disable it in the LTI, until we have time to go through the Sentry configurations to make sure we aren't in violation.

hedera/urls.py
dodget: Sure thing, I'll changed that up.

lti/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    It expects there to be lis_person_contact_email_primary in the POST.
+    Returns the login function.
+    """
+
+    lti_user_email = request.POST.get("lis_person_contact_email_primary", None)
+    if lti_user_email is None:
+        return False
+    account_email = EmailAddress.objects.get(email=lti_user_email)
+    user = get_user_model().objects.get(username=account_email.user.username)
+    login(request, user, backend="hedera.backends.UsernameAuthenticationBackend")
```
paltman, line 18: Cut two unnecessary queries by using `select_related` and then the relation off the `EmailAddress` object.

```suggestion
    account_email = EmailAddress.objects.filter(email=lti_user_email).select_related("user").get()
    login(request, account_email.user, backend="hedera.backends.UsernameAuthenticationBackend")
```

lti/views.py
```diff
@@ -0,0 +1,183 @@
...
+                lti_user = login_existing_user(request)
+            except EmailAddress.DoesNotExist:
+                request.session["lti_email"] = request.POST.get(LTI_PROPERTY_USER_EMAIL, None)
+                if request.session["lti_email"] is None:
+                    return render(request, "lti_failure.html")
+                return HttpResponseRedirect(reverse("lti_registration"))
+            if lti_user is False:
+                return render(request, "lti_failure.html")
+
+        return super(LtiInitializerView, self).dispatch(request, *args, **kwargs)
```
paltman, line 59: You no longer need to provide current class and `self` when calling `super()`.

```suggestion
        return super().dispatch(request, *args, **kwargs)
```

lti/views.py
```diff
@@ -0,0 +1,183 @@
...
+        lti_params = {
+            "course_id": request.session.get(LTI_PROPERTY_COURSE_ID, None),
+            "roles": request.session.get(LTI_PROPERTY_USER_ROLES, None),
+            "title": request.session.get(LTI_PROPERTY_COURSE_TITLE, None)
+        }
+
+        if None in lti_params.values():
+            return render(request, "lti_failure.html")
+
+        self.initialize_group(
```
paltman, line 73: It's generally bad practice to mutate anything on `GET` requests.  Is this necessary to do here?  It's not clear to me reading through things what the goal of calling `GET /lti/lti_initializer/` goal is or why it can't be a `POST` if it needs to create/change data.

lti/views.py
```diff
@@ -0,0 +1,183 @@
...
+    def get_or_create_group(self, course_id=None, title="Empty", user=None):
+        """ takes (int) course_id, and (str) title and returns a Group"""
+        try:
+            group = Group.objects.get(class_key=course_id)
+        except Group.DoesNotExist:
+            group = Group.objects.create(
+                class_key=course_id,
+                title=title,
+                created_by=user
+            )
```
paltman, line 136: Let's use `get_or_create` with `defaults`:

```suggestion
        group = Group.objects.get_or_create(
            class_key=course_id,
            defaults=dict(
                title=title,
                created_by=user
            )
        )
```

lti/views.py
```diff
@@ -0,0 +1,183 @@
...
+            if user in group.teachers.all():
+                group.teachers.remove(user)
+        return None
+
+
+class LtiRegistrationView(FormView):
+
+    template_name = "lti_registration.html"
+    form_class = LtiUsernameForm
+    success_url = "/lti/lti_initializer/"
```
paltman, line 160: By using `reverse` we can avoid hard coding the URL and keep things DRY:

```suggestion
    success_url = reverse("lti_initializer")
```

lti/views.py
```diff
@@ -0,0 +1,183 @@
...
+        try:
+            request.session["lti_email"]
+        except KeyError:
+            return render(request, "lti_failure.html")
+        return super(LtiRegistrationView, self).dispatch(request, *args, **kwargs)
+
+    def form_valid(self, form):
+
+        username = form.cleaned_data["username"]
+        unique_username = form.unique_username(username)
```
paltman, line 172: This should be done as part of Django's built in form validation and then handle the invalid case by implementing a `form_invalid` method on this view.

lti/views.py
```diff
@@ -0,0 +1,183 @@
...
+        if unique_username is False:
+            msg = "Sorry, that username is taken."
+            form.add_error("username", msg)
+            return super().form_invalid(form)
+        email = self.request.session["lti_email"]
+        password = get_random_alphanumeric_string(10)
+
+        new_user = get_user_model().objects.create_user(username, email=email, password=password)
+        login(self.request, new_user, backend="hedera.backends.UsernameAuthenticationBackend")
```
paltman, line 181: Need to make sure other related objects for the user are created like `Account` and `EmailAddress`.

hedera/settings.py
paltman: By default PII should be turned off with latest Sentry SDKs which should satisfy FERPA.  We use Sentry in software that we've built in FERPA environments.  But by all means, review this https://docs.sentry.io/data-management/sensitive-data/ and decide for yourself.

lti/views.py
dodget: Good question. It was to handle the `success_url` redirect from the username registration form. I've never attempted to specify that redirection come in the form of `POST`, so I was just grabbing the lti properties that were set in the session by `django-lti-provider` in order to initialize the user's groups and roles.

Since Hedera moved to using a display_name, there is no need for a username registration form now, so this all is getting deleted anyways.

lti/views.py
dodget: From my observations, `Account` and `EmailAddress` get created during this process. There is a test to verify that an `Account` is created, but I should probably include `EmailAddress` in that test.

lti/views.py
dodget: 👍  I wasn't sure how `defaults` worked with `get_or_create`

lti/views.py
dodget: This view has been removed

lti/views.py
dodget: This form has been removed.

lti/views.py
dodget: 👍

lti/utils.py
dodget: 👍

dodget:
There has been some refactoring to this branch since the inclusion of user profile and the `display_name` attribute. To summarize changes from the original PR:

- All username registration (LtiUsernameForm, LtiRegistrationView, template) code has been removed
- Adding a `display_name` during user creation has been included, which uses the username portion of the user's email address
- Since the username registration form was removed, users should never need to `GET /lti/llti_initializer/`, so the `get` method of `LtiInitializerView` has been removed.
- Allow lti users to access the user settings so they can change their display name etc, but restrict them from changing their email address.

Notes:
- I've included all the suggestions for code-improvement
- LTI users can now access part of the user settings where they can change their password. That wouldn't have a use for those users, and may generate confusion. I would hide it from lti users if possible. The user settings template -- I believe -- is coming from the pinax django-user-accounts package, so making a change in this PR doesn't seem worth the effort.

@joshuagetega, @arthurian, @paltman

lti/tests.py
```diff
@@ -0,0 +1,144 @@
...
+    def test_dispatch(self, mock_validate_request):
+        """ Test successful dispatch redirects to home """
+        mock_validate_request.return_value = True
+        response = self.client.post(
+            "/lti/lti_initializer/",
+            {
+                "lis_person_contact_email_primary": "user1@example.com",
+                "custom_canvas_course_id": "327",
+                "context_title": "test title",
+                "ext_roles": "urn:something/something-else/Instructor,urn:something/something-else/Student"
```
arthurian, line 125: It might be good to use some realistic values for `ext_roles` to make the test closer to what will actually be sent by the LTI consumer (e.g. Canvas). Here are a few real examples that could be used:

1. As an Instructor in a course (`urn:lti:role:ims/lis/Instructor`):
```
ext_roles=urn:lti:instrole:ims/lis/Instructor,urn:lti:instrole:ims/lis/Student,urn:lti:role:ims/lis/Instructor,urn:lti:sysrole:ims/lis/User
```

2. As a TeachingAssistance in a course (`urn:lti:role:ims/lis/TeachingAssistant`):
```
ext_roles=urn:lti:instrole:ims/lis/Instructor,urn:lti:instrole:ims/lis/Student,urn:lti:role:ims/lis/TeachingAssistant,urn:lti:sysrole:ims/lis/User
```

3. As a Student in a course (`urn:lti:role:ims/lis/Learner`):
```
ext_roles=urn:lti:instrole:ims/lis/Student,urn:lti:role:ims/lis/Learner,urn:lti:sysrole:ims/lis/User
```

joshuagetega (approved):
Looks good, @dodget. Great PR.

paltman:
Is this ready for merge?  I'm not sure if based on most recent comments (25 days ago) if this is done and just pending a final review by one of us or not.  I'll give this another review either way as I don't think it could hurt but at some point I think we bite the bullet and get this merged in either way.

hedera/settings.py
paltman: @dodget what do you think about turning this back on given the default settings for Sentry and PII?

paltman (approved):

PR #236, opened 2020-08-19
Font support for cyrillic and combining diacritics

arthurian:
This PR fixes Russian text that includes combining diacritics (particularly acute accents). This was previously working with the google font CSS API v1, but it appears to have broken when we switched to CSS API v2.

Example:

|Broken|Fixed|
|-|-|
|broken-text|fixed-text|

PR changes:

1. Modifies the [Noto Serif](https://fonts.google.com/specimen/Noto+Serif) font request so it includes the correct subset for Russian (changed `subset=russian` to `subset=cyrillic`).
2. Appends the `text` parameter to the font request with a list of required characters and diacritics that need to be included (e.g. 33 characters of russian alphabet and some combining diacritics). For reference, the diacritics that are URL encoded include: 
    - 0301 COMBINING ACUTE ACCENT
    - 033D COMBINING X ABOVE
    - 0308 COMBINING DIAERESIS
    - 0306 COMBINING BREVE

It's unclear why combining diacritics are not being rendered properly even with `subset=cyrillic` or `subset=cyrillic-ext`. Serving the Noto font locally does appear to resolve the issue, so it's not an issue with the font itself, but rather with the [google font API v2](https://developers.google.com/fonts/docs/css2). We could consider hosting the Noto font ourselves instead of using the google font API, since it would insulate us from any changes and would resolve this issue in a cleaner way, but I hesitate to do that in case there's a simpler solution.

---

**UPDATE:** revised this PR based on @dodget's feedback so that the Noto font (serif and sans-serif versions) are served by the app itself. This resolves the issue.


dodget:
I'd be in favor of hosting the font ourselves. Seems like a logical solution, especially since stable/consistent font behavior should probably be a requirement for the project. What is one more static asset in s3?

arthurian:
That's a very good point @dodget. I'll revise this PR using a self-hosted version of the Noto Serif font.

joshuagetega (approved):
This looks good, @arthurian.

arthurian:
@bilbe @paltman Any objections to merging this PR?

PR #240, opened 2020-08-26
Tidy up static handling

paltman:
I noticed there was a second static folder under hedera/ that contained favicons, fonts, and a css file for the fonts.

I also noticed we were still using Django's static tag to load those.  Combined everything into our single webpack build.


paltman:
@joshuagetega I'm going ahead and merging this but wanted your eyes on this in case in impact deployments, I don't think it should but just in case.

PR #266, opened 2020-10-13
Learners can print/save a PDF handout of the text with glosses

paltman:
Screen Recording 2020-10-12 at 07 39 59 PM


paltman:
This is just some rudimentary style proving out the idea of a printable handout.  The second tab is a PDF generated and displayed in browser.  We can flip it easily to trigger a download.  Whatever you prefer.

It's currently not pulling in our site fonts and that might be a limitation of the backend service we are using.  But I also need to do some styling on the page breaks.  But I'm also open to any styling/layout ideas you might have as well so wanted you to see it sooner than later, @IvyJL.

IvyJL:
Ooo! Cool! We may have to do some research on whether people prefer alphabetical order or order of first occurrence. I also wonder if a side-by-side view is now more popular.

paltman:
Here is an updated sample:

[document-1.pdf](https://github.com/Hedera-Lang-Learn/hedera/files/5380595/document-1.pdf)

IvyJL:
That looks nice! If we go with side-by-side, then I think it perhaps should go in order of appearance. Otherwise there will be endless flipping.

paltman:
[document(2).pdf](https://github.com/Hedera-Lang-Learn/hedera/files/5422822/document.2.pdf)

IvyJL:
This looks nice! Ideally, the learner won’t need as many glosses as this ;)

joshuagetega (approved):
Looks good. Successfully tested it locally.

PR #270, opened 2020-10-15
moving dev env into docker

paltman:
I recently destroyed my local development environment with a `brew install` mistake that left me with cross versions of Redis and Postgres.  I think moving to Docker is a general improvement especially in the context of a larger team because now the versions of all the services used are locked down and explicit as described in the `docker-compose.yml` file.

I've done my best to make this additive (left Pipenv for example) so as to be an opt-in for others for local dev.  You shouldn't be _required_ to do anything with Docker and it should not interrupt how you currently develop.  If it does, please let me know.


paltman:
@joshuagetega please review and merge if you are ok with what I've done here

joshuagetega (approved):
Looks good, @paltman. I will go ahead and merge.

PR #285, opened 2021-01-28
Disable SESSION_COOKIE_SECURE in local dev

jacobwegner:
Default the setting to False, but set the default for the
SESSION_COOKIE_SECURE  env var in web-entrypoint.sh to 1.

This enables a developer working locally to use http cookies,
but still enforces https cookies in prod-like environments.


joshuagetega (approved):
Looks good!

PR #291, opened 2021-04-07
Siteimprove: Accessibility Changes

bilbe:
This is the first round of changes for the sake of meeting accessibility standards per Siteimprove. These have mostly focused on images but also an HTML tag suggestion to use STRONG instead of BOLD. There will be more changes coming. Some suggestions I will ignore, e.g., Siteimprove suggest that our home page is too difficult to read because apparently you need to be at least college sophomore to understand those 25 words (5 of which are polysyllabic!)... sad.


hedera/templates/cms/home_page.html
```diff
@@ -11,7 +11,7 @@
 
 {% block body_base %}
 
-  <section class="home-hero">
+  <section class="home-hero" title="Hedera Logo">
```
jtauber, line 14: I'd maybe title it "Hedera Header" instead as it's really the whole header section, not just the logo.

jtauber (dismissed):
was `static/src/images/site_base.html` checked in by mistake?

otherwise, just one small title change suggestion

static/src/js/app/components/DownloadVocab.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <a class="btn btn-sm btn-light" :href="glossesDownload" download="glosses.csv">
-    <icon name="download" /> Export
+    <i class="fa fa-fw fa-download" aria-hidden="true" /> Export
```
jtauber, line 3: converting <icon> to <i> might cause reactivity problems

hedera/templates/cms/home_page.html
bilbe: done.

static/src/js/app/components/DownloadVocab.vue
bilbe: I had a good look and I don't think this is true. It looks like the few places that there are interactive icons they are done through css shading. I think we're good.

PR #295, opened 2021-05-19
updated packages in package.json to newer version

Tanvez:


Tanvez:
> Probably want to include the lock file too.

Just added!

jtauber (approved):

Issue #297, opened 2021-05-28
As a user, I'd like to have a dashboard landing view


arthurian:

arthurian:
Acceptance Criteria:
- After user logs in, they should be redirected to the dashboard instead of the home page.
- The dashboard should be at a different URL than the home page so that you can still return to the home page.

jtauber:
The dashboard should only be accessible if logged in. (It's meaningless if you're not).

PR #312, opened 2021-06-14
Redirect login dashboard

Tanvez:
Relates to issue #297 
- Added dashboard.html placeholder to redirect logged in users to a dashboard view.
- Added tests folder and test_views.py to test dashboard renders and `redirect_chain`
- Added `LoginRequiredMixins` (thanks for the suggestion Tylor!)
Note: 
- I used TemplateView, let me know if theres another way/best practice to implement this instead.
- I also hard coded the redirect url into the test


hedera/views.py
```diff
@@ -54,3 +55,7 @@ def after_signup(self, form):
         profile = self.created_user.profile
         profile.display_name = form.cleaned_data["display_name"]
         profile.save()
+
+
+class DashboardView(TemplateView):
+    template_name = "dashboard.html"
```
dodget, line 61: Since this is so simple, I support including the dashboard view in `hedera.views` at this point. But if if it starts to get more complicated, we could consider moving it to its own app.

hedera/urls.py
```diff
@@ -22,6 +22,7 @@
     path("django-rq/", include("django_rq.urls")),
     path("account/login/", LoginView.as_view(form_class=LoginEmailForm), name="account_login"),
     path("account/signup/", views.SignupView.as_view(), name="account_signup"),
+    path("account/dashboard/", views.DashboardView.as_view(), name="dashboard"),
```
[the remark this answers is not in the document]
jtauber: I would also suggest just putting it as a top-level `/dashboard/` rather than under `/account/`. This is the pattern we've always followed with Pinax sites using django-user-accounts, etc.

hedera/views.py
jtauber: I wonder though if this should be a authenticated-only view. If someone hits `/dashboard/` unauth'd we probably want to redirect them to login rather than give a dashboard with silent errors because `request.user` is not a proper User.

hedera/urls.py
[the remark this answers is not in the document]
Tanvez: Just to clarify, can the dashboard view be accessible when the user is not logged in? I was working off the specs for issue #297, would definitely update the paths if the specs changed. let me know!

hedera/views.py
dodget: Yeah, the url path should be wrapped in `login_required`

hedera/urls.py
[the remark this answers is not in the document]
dodget: I don't think the specs are changed. The user should be logged in to get to the dashboard, but that doesn't mean the url needs `account/` in it. To update my earlier suggestion, the path could be:
`path("dashboard/", login_required(views.DashboardView.as_view()), name="dashboard"),`
But you'd have to import `login_required` from the appropriate place.

hedera/views.py
Tanvez: Makes sense! Will update the PR

hedera/urls.py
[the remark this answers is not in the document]
Tanvez: awesome thanks for the clarification @dodget !

hedera/settings.py
```diff
@@ -330,6 +330,7 @@
 ACCOUNT_EMAIL_CONFIRMATION_REQUIRED = False
 ACCOUNT_EMAIL_CONFIRMATION_EXPIRE_DAYS = 7
 ACCOUNT_USE_AUTH_AUTHENTICATE = True
+ACCOUNT_LOGIN_REDIRECT_URL = "/dashboard"
```
arthurian, line 333: This setting handles the redirect on login, but what about signup? Currently, after you sign-up, it redirects you back to the home page, but I'm assuming we'll want to redirect to the dashboard in that case as well. 

It looks like there's a separate [ACCOUNT_SIGNUP_REDIRECT_URL](https://github.com/pinax/django-user-accounts/blob/68bac737dbcd1f5f407aae80d4783b496fcfded8/docs/settings.rst#account_signup_redirect_url) setting that we would need to add.

arthurian (approved):
Looks good @Tanvez! Just two minor suggestions.

hedera/settings.py
Tanvez: Oh I didnt think about that! Thanks Artie will add that to the code.

hedera/views.py
```diff
@@ -54,3 +57,10 @@ def after_signup(self, form):
         profile = self.created_user.profile
         profile.display_name = form.cleaned_data["display_name"]
         profile.save()
+
+
+# https://docs.djangoproject.com/en/3.2/topics/auth/default/#the-loginrequired-mixin
+class DashboardView(LoginRequiredMixin, TemplateView):
+
+    login_url = "/account/login/"
```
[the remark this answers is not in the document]
Tanvez: Good call, Will remove that since its not need

Issue #310, opened 2021-06-03
As a teacher, I want to be able edit a text so that I can fix mistakes, etc


arthurian:
Possible Approaches:
- Custom editor (discussed simple add/delete/replace type editor with hotkeys)
- Diffing

Spike to explore editing the text without losing the existing lemmatization/disambiguation.

Related to: #308

bilbe:
I believe this was already done by Tylor

PR #313, opened 2021-06-17
Cloning texts

### LINKED ISSUE #308: As a user, I'd like to be able to clone texts so that the lemmatization is preserved
Ability to copy the text + lemmatization (currently just copies the text, losing the hard-work of disambiguating the text).
Perhaps all copies should be like point a, but then we need a mechanism for editing the text that doesn’t lose lemmatization. (e.g., substitute: interficio for occido)

dodget:
This PR resolves #308 and [ATGU-3108](https://jira.huit.harvard.edu/browse/ATGU-3108) by:

- Updating `LemmatizedText`'s `clone` method based on django's recommended process: https://docs.djangoproject.com/en/3.2/topics/db/queries/#copying-model-instances
- Adding a test to verify cloning functionality
- Editing the `create` view for the new (non-edited) cloning process

## Notes
- This functionality appears to be worthless with out the ability to edit lemmatized texts (as in #310 and #298)
- This functionality would be valuable if there was an `Edit` button next to `Delete` and `Clone` in the lemmatized texts list
- Since cloned texts are no longer dropped into the editor, I removed all the now ununsed `cloned_from` context data from the view and the template.


arthurian (approved):
Looks good @dodget, I just have some questions about the the fields that are being cloned.

lemmatized_text/tests.py
```diff
@@ -0,0 +1,39 @@
...
+            created_by=self.created_user1,
+            data=json.dumps({"key": "test"})
+        )
+        self.assertEqual(LemmatizedText.objects.all().count(), 1)
+        lemmatized_text.clone()
+        self.assertEqual(LemmatizedText.objects.all().count(), 2)
+        lemmatized_text.clone(cloned_by=self.created_user2)
+        self.assertEqual(LemmatizedText.objects.all().count(), 3)
+        self.assertEqual(LemmatizedText.objects.filter(created_by=self.created_user1).count(), 2)
+        self.assertEqual(LemmatizedText.objects.filter(created_by=self.created_user2).count(), 1)
```
Tanvez, line 39: Just an observation but should we change the title or have an indicator that the text is a clone?
Everything Looks great Tylor!

Tanvez (approved):

lemmatized_text/models.py
```diff
@@ -146,14 +146,14 @@ def clone_url(self):
...
-            data=self.data,
-            created_by=cloned_by or self.created_by,
-        )
+        """Set a copy of this object's pk to None, set some relationships and save (cloning)"""
+        # https://docs.djangoproject.com/en/3.2/topics/db/queries/#copying-model-instances
+        obj = LemmatizedText.objects.get(pk=self.pk)  # declaring 'self' here results in a ValueError
+        obj.pk = None
+        obj.cloned_from = self
+        obj.created_by = cloned_by or self.created_by
+        obj.save()
```
Tanvez, line 155: Just an observation but should we change the title or have an indicator that the text is a clone?
Everything Looks great Tylor!

dodget: I think it's a good idea, but I hesitate to programmatically edit any user provided content. But there should be some way to more easily distinguish between original text and cloned text. I think the best UX will shake out after some use.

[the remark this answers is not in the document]
dodget: Will do on the `secret_id`. I think `created_at` should probably be the cloned date, which makes me think I need to set it.

PR #316, opened 2021-06-21
Update readme

Tanvez:
Updated the README.md with Contributing Guidelines sections which has local environment linting instructions so that code is linted before pushing up/PR

The instructions cover `isort`, `flake8` and `eslint` commands for Python and Javascript respectively.

I will also add a diagram for the project structure in the WIKI (WIP)

Note: I know that theres a way for pre-commit(can run linters) to run before code is committed upstream for Node/Javascript but im not sure about Django/Python. Can look into it more if we want to add something like this down the line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,3 +116,32 @@ For local development:
 ```
 docker run --publish 9000:9000 --env X_API_KEY=test-secret eldarioninc/pdf-service
 ```
+
+## Contributing Guidelines
+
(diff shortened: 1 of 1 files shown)


dodget (approved):
LGTM, thanks Vesna!

arthurian (approved):
Thanks @Tanvez!

Issue #302, opened 2021-05-28
As a user, I want to be able to quickly add a word to my vocab list from my dashboard


arthurian:

arthurian:
Acceptance criteria:
- Dashboard should provide a form to add a new word to the user's vocab list.
- Enter a word, then return possible lemmas that match the word.
- User should pick one of the possible lemmas.
- User should accept the existing gloss OR override the gloss with their own.
- If the word is already in their personal vocab list, show that to the user (e.g. one of the lemmas is in their list, and others are not).

Notes:
- Two separate issues - linking to the lattice, and the second is adding the gloss you want to use.
- What happens if user tries to enter a word that has no associated lemma node? Punt on this issue for now (handle separately since we don't have a good way of dealing with it right now).

bilbe:
Look at Import Vocab: https://dev.hederaproject.org/vocab/personal/create/

Separate word and definition fields with an auto select list on the right side:
![image](https://user-images.githubusercontent.com/5712116/121557936-2f1b2780-c9e3-11eb-9ca9-0384129b741d.png)

Grab the ID to connect new word up. When word is chosen from list then fill in definition but allow user to change both word and definition as well as select comfort level:

![image](https://user-images.githubusercontent.com/5712116/121557777-06932d80-c9e3-11eb-91f4-430b0a16cdd3.png)

PR #321, opened 2021-07-07
Quick add vocab

Tanvez:
Related to #302 
- Added a Vue modal which brings up a form to add a headword, gloss and familiarity based on the Vocabulary List that are already present in the database
- Created api endpoints to get list of personal vocab languages and to add vocab based on the list the user has selected
- Added python test for these new endpoints
- Added Vue Jest test suite which generate coverage reporting for code coverage via `npm run test:unit`
- Added basic test for QuickAddVocabForm, Mutations, and Actions relating to the form
Note: i commented out some expect lines because it was not returning what i expected Example: `expect(commit).toHaveBeenCalledWith('createPersonalVocabEntry', response)`  it would instead return `('createPersonalVocabEntry', undefined)`  Not sure why that is, would have to look more into it
- Followed https://accessibility.huit.harvard.edu/technique-accessible-modal-dialogs for modal accessibility


hedera/api.py
```diff
@@ -293,3 +297,35 @@ def post(self, request, *args, **kwargs):
         vl.refresh_from_db()
 
         return JsonResponse({"data": vl.data()})
+
+
+class PersonalVocabularyLangListAPI(APIView):
+
+    def get_data(self):
+        qs = PersonalVocabularyList.objects.filter(user=self.request.user)
+        data = serializers.serialize('json', qs)
```
Tanvez, line 306: look into removing serializer since we are iterating

hedera/urls.py
```diff
@@ -47,6 +47,9 @@
     path("api/v1/vocab_entries/<int:pk>/<str:action>/", api.VocabularyListEntryAPI.as_view()),
     path("api/v1/personal_vocab_list/", api.PersonalVocabularyListAPI.as_view()),
     path("api/v1/personal_vocab_list/<int:pk>/", api.PersonalVocabularyListAPI.as_view()),
+    path("api/v1/personal_vocab_list/list_lang/", api.PersonalVocabularyLangListAPI.as_view()),
+    path("api/v1/personal_vocab_list/add/", api.PersonalVocabularyLangListAPI.as_view()),
```
Tanvez, line 51: make into one path

static/src/js/app/constants.js
```diff
@@ -27,3 +31,8 @@ export const RATINGS = {
   4: 'I definitely know what this word means but could forget soon',
   5: 'I know this word so well, I am unlikely to ever forget it',
 };
+
+export const LANGUAGES = {
+  lat: 'Latin',
+  grc: 'Greek',
```
Tanvez, line 37: add russian

static/src/js/app/components/QuickAddVocabModal.vue
```diff
@@ -0,0 +1,168 @@
...
+                    v-model="headword"
+                    required
+                  />
+                  <input
+                    class="form-control ml-2 mt-2"
+                    type="text"
+                    placeholder="gloss"
+                    v-model="gloss"
+                    required
+                  />
```
dodget, line 57: One aspect of accessible forms is that fields need labels (ideally with semantic html). Placeholder attributes can serve that purpose for folks who do not have vision issues, but I don't think they do much good for screen reader application.

Tanvez: Thanks Tylor! Will look into semantic html for accessibility and what is available for screen reader compliance.

dodget (approved):
Looks good, this was a big lift!

static/src/js/app/constants.js
arthurian: @Tanvez Can you add `rus: 'Russian'` to the languages for now, so that it matches the `SUPPORTED_LANGUAGES` in `hedera/settings.py`? Eventually we can add an endpoint for that, but just want to make sure it matches for now.

arthurian (approved):
Looks good! Great work @Tanvez :+1:

static/src/js/app/constants.js
arthurian: Nevermind, I just missed the fact that you already added this. My bad!

PR #324, opened 2021-07-15
Enhancements to Quick Add Personal Vocab Feature

Tanvez:
[ATGU-3145](https://jira.huit.harvard.edu/browse/ATGU-3145) and [ATGU-3161](https://jira.huit.harvard.edu/browse/ATGU-3161)
- added select options for related nodes in modal
- added test for django and vue
- added preferred language
- make node selection accessible via arrow keys

Screen Recording 2021-07-29 at 2 07 25 PM

Update:
- refactor lattice_nodes api to look at lemma_nodes when building the lattice_nodes list relating to a headword
- refactor regex to include trailing numbers in results
- updated vue select options for suggested gloss to strip out numbers from lattice_node labels
- updated django test in relation to updated api


dodget (approved):
Great work Vesna! I'd like to answer a couple of questions before merging, mainly around adding the preferred language to the user profile.

lemmatized_text/migrations/0015_alter_lemmatizedtext_cloned_for.py
```diff
@@ -0,0 +1,20 @@
...
+        ('lemmatized_text', '0014_auto_20201012_2328'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='lemmatizedtext',
+            name='cloned_for',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, to='groups.group'),
+        ),
+    ]
```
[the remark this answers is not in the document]
dodget: This rogue migration could be the result of django drift. Probably nothing that you did.

hedera/api.py
```diff
@@ -312,17 +319,20 @@ def get_data(self):
...
+
+    def get_data(self):
+        headword = self.request.GET.get("headword")
+        filtered_headword_iterable = filter(str.isalnum, headword)
+        filtered_headword_string = "".join(filtered_headword_iterable)
+        qs = LatticeNode.objects.filter(label__iregex=r"\y" + re.escape(filtered_headword_string) + r"\y")
+        lattice_node_list = []
+        for node in qs:
+            lattice_node_list.append(node.to_dict())
+        return lattice_node_list
```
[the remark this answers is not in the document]
Tanvez: oh I didn't know that! Thats a cool feature! Will update the code with those changes.

hedera/migrations/0004_profile_lang.py
```diff
@@ -0,0 +1,18 @@
...
+        ('hedera', '0003_profile_show_node_ids'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='profile',
+            name='lang',
+            field=models.CharField(max_length=3, null=True),
+        ),
+    ]
```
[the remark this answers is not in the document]
Tanvez: Yea, definitely! Will hold off on any merging until 👍 by Bill.

hedera/tests/test_api.py
```diff
@@ -20,8 +25,45 @@ def test_post_personal_vocabulary_quick_add_api(self):
...
+class MeAPITest(APITestCase):
+    def setUp(self):
+        self.created_user = User.objects.create_user(username=f"test_user{randrange(100)}", email=f"test_user{randrange(100)}@test.com", password="password")
+
+    def test_post_me_profile_with_lang_field(self):
+
+        self.client.force_login(user=self.created_user)
+        response = self.client.post("/api/v1/me/", json.dumps({"lang": "lat"}), content_type="application/json")
+        data = response.json()["data"]
+        self.assertEqual(data["lang"], "lat")
```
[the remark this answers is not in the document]
Tanvez: ahh will update the test with failing tests!

lemmatized_text/migrations/0015_alter_lemmatizedtext_cloned_for.py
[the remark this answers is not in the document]
Tanvez: It was generated when I ran the migrations for the Profile model, not sure if it was a previous change that was never migrated or django drift 🤔

hedera/tests/test_api.py
```diff
@@ -54,16 +94,27 @@ def test_get_related_lattice_nodes(self):
...
+
+    def test_successful_post_me_profile_with_lang_field(self):
         response = self.client.post("/api/v1/me/", json.dumps({"lang": "lat"}), content_type="application/json")
         data = response.json()["data"]
         self.assertEqual(data["lang"], "lat")
+
+    def test_failing_post_me_profile_with_lang_field(self):
+        # if language is not supported
+        response = self.client.post("/api/v1/me/", json.dumps({"lang": "abc"}), content_type="application/json")
+        self.assertEqual(response.status_code, 400)
```
dodget, line 120: 👍

static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
```diff
@@ -0,0 +1,275 @@
...
+  } from '../../constants';
+  import FamiliarityRating from '../../modules/FamiliarityRating.vue';
+
+  export default {
+    components: { FamiliarityRating },
+    // on creation of the dom element fetch the list of langauages/ids the user has in their personal vocab list
+    async created() {
+      await this.$store.dispatch(FETCH_PERSONAL_VOCAB_LANG_LIST);
+      await this.$store.dispatch(FETCH_ME);
+      window.addEventListener('keyup', this.onKey);
```
Tanvez, line 127: todo: remove dead code

hedera/api.py
```diff
@@ -310,19 +320,29 @@ def get_data(self):
...
-#         headword = self.request.GET.get("headword")
-#         filtered_headword_iterable = filter(str.isalnum, headword)
-#         filtered_headword_string = "".join(filtered_headword_iterable)
-#         qs = LatticeNode.objects.filter(label__iregex=r"\y"+ re.escape(filtered_headword_string) + r"\y")
-#         data = serializers.serialize('json', qs)
-#         json_data = json.loads(data)
-#         return json_data
+        checked_data = self.check_data(data)
+        if checked_data is not True:
+            return HttpResponseBadRequest("Missing required fields")
```
arthurian, line 334: Should we return these 400 responses as JSON, that way the client knows the errors will be returned as JSON as well? Maybe something like this (just a suggestion):

```python
class JsonResponseBadRequest(JsonResponse):
    status_code = 400
```

And then you could write:

```python
return JsonResponseBadRequest(data={"error": "Missing required fields"})
```

hedera/api.py
```diff
@@ -310,19 +320,29 @@ def get_data(self):
...
+        return JsonResponse({"data": {"created": True, "data": new_entry.data()}})
+
+
+class LatticeNodesAPI(APIView):
+
+    def get_data(self):
+        headword = self.request.GET.get("headword")
+        filtered_headword_iterable = filter(str.isalnum, headword)
+        filtered_headword_string = "".join(filtered_headword_iterable)
+        qs = LatticeNode.objects.filter(label__iregex=r"\y" + re.escape(filtered_headword_string) + r"\y")
```
arthurian, line 347: Just curious, but what is `\y` in the regex?

hedera/migrations/0004_profile_lang.py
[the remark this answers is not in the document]
arthurian: Agreed on getting input from @bilbe.

lemmatized_text/migrations/0015_alter_lemmatizedtext_cloned_for.py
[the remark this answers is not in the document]
arthurian: Yeah, I suspect this is drift. And as we discussed offline, we should pin it at django 3.2 now so that doesn't happen again (at least not without our knowing).

static/src/js/app/Dashboard.vue
```diff
@@ -1,28 +1,13 @@
...
-    <Modal
-      ref="Modal"
-      title="Add Word to Personal Vocabulary List"
-      description="Modal to quick add vocab"
-      ><QuickAddVocabForm ref="QuickAddVocabForm"
-    /></Modal>
-    <button type="button" class="btn btn-primary" @click="openModal">
-      Quick Add Vocab
-    </button>
+    UNDER CONSTRUCTION
```
arthurian, line 3: I suggest we remove this, since we can assume it's under construction.

static/src/js/app/components/quick-add-button/QuickAddButton.vue
```diff
@@ -0,0 +1,32 @@
...
+</template>
+<script>
+  import Modal from '../Modal.vue';
+  import QuickAddVocabForm from './QuickAddVocabForm.vue';
+  import { FETCH_ME } from '../../constants';
+
+  export default {
+    components: { Modal, QuickAddVocabForm },
+    async created() {
+      await this.$store.dispatch(FETCH_ME);
```
arthurian, line 22: I'm wondering if it would make sense to have the `FETCH_ME` action dispatched from the `Dashboard.vue` component instead of here. My assumption is that the "me" data only needs to be read once, when the dashboard is loaded. I could imagine other sub-components on the dashboard needing that data, and we wouldn't want every sub-component dispatching the same action. So if the dashboard managed that state by dispatching the action, then sub-components wouldn't need to manage it, they could just read the state directly from the store. I could be thinking about this wrong, so curious what you think.

static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
```diff
@@ -0,0 +1,275 @@
...
+          headword: this.headword,
+        });
+        // sets first latticenode as the default in select options
+        if (this.latticeNode) {
+          this.latticeNodeId = this.latticeNode[0].pk;
+        }
+      },
+      onSelect(event) {
+        // note: '+' changes the string to a number
+        const node = this.latticeNode.find((ele) => +ele.pk === +event.target.value);
```
arthurian, line 221: 👍  for the note about the `+`. That's a neat trick, although you could make the case that `parseInt()` is clearer and reflects the expected format of the number.

arthurian (approved):
Nice work, @Tanvez! It's great to see how the QuickAdd components are evolving, along with the backend and frontend unit tests. Kudos for making those better.

Overall this looks like it's in good shape to merge, other than a few questions around the preferred language that @dodget mentioned. I added a few questions/suggestions, but nothing that should block you from moving forward.

hedera/api.py
Tanvez: Yea, we definitely should! Will update the code to reflect this.

hedera/api.py
Tanvez: Think it might have something to do with how Postgres uses word boundaries. I found an old stackoverflow post here [PostgreSQL Regex Word Boundaries](https://stackoverflow.com/questions/3825676/postgresql-regex-word-boundaries/29798772). We might have to tweak this to accommodate the changes Ivy was talking about.

static/src/js/app/Dashboard.vue
Tanvez: Will do!

static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
Tanvez: ahh that would be more descriptive! will update with the `parseInt()`

hedera/api.py
Tanvez: Should we hold off on refactoring the regex or refactor the regex to account for numerals in the headword?

static/src/js/app/components/quick-add-button/QuickAddButton.vue
Tanvez: I think due to how hedera loads in different vue components periodically some of the vuex state might not persist between page navigation. Since we are planning to add this button site wide, might need to be updated in [ATGU-3160](https://jira.huit.harvard.edu/browse/ATGU-3160)

arthurian: Yeah, it may require further refactoring when we get to that story.

hedera/api.py
Tanvez: @jtauber We had a question about the following when looking up headwords in relation to the personal vocabulary list:
- if we wanted to search for a headword should we look at `lattices_latticenodes`  table
- how are the headword stored based on multiple meanings between the same headword

hedera/api.py
```diff
@@ -393,8 +393,34 @@ def get_data(self):
...
+        lattice_node_list = []
+        for node in lattice_nodes:
+            gloss = node["gloss"]
+            if gloss != "from morpheus" and gloss != "morpheus ambiguity":
+                lattice_node_list.append(node)
+            elif len(node["children"]):
+                # checks child nodes with matching label and appends them to the result
+                for child_node in node["children"]:
+                    valid_lemma = bool(re.search(rf"\b{headword}[0-9]*\b", child_node["label"]))
+                    if child_node["gloss"] != "from morpheus" and child_node["gloss"] != "morpheus ambiguity" and valid_lemma is not False:
```
Tanvez, line 411: todo: leave off check for valid_lemma

PR #325, opened 2021-07-19
fix flake8 precommit to check double quotes

Tanvez:
Quick fix for flake8
- flake8 is not picking up the configuration setting for `inline-quotes = 'double'` so I added it in the args field in pre-commit-config.yaml


arthurian (approved):
LGTM! 👍

dodget (approved):
LGTM

PR #327, opened 2021-07-26
Django cke

dodget:
This PR resolves [ATGU-3155](https://jira.huit.harvard.edu/browse/ATGU-3155) and #310 by:

- Adding the `django-ckeditor` package
- Configuring the ckeditor in settings.py
- Adding an “Edit” lemmatized text form and template
- Adding an `edit` lemmatized text view to handle the form GET and POST
- Adding `LemmatizedText` methods to help with transforming data to html
- Adding `LemmatizedText` methods to handle an edited text
- Adding an `EditedTextHtmlParser` to parse html back to the lemmatized text data format
- Adding a `TagStripper` parser to update the text’s “original text” data
- Adding tests for views and parsing


lemmatized_text/parsers.py
```diff
@@ -0,0 +1,122 @@
+from html.parser import HTMLParser
+from io import StringIO
+
+from lemmatization.lemmatizer import Lemmatizer
+
+
+class EditedTextHtmlParser(HTMLParser):
```
Tanvez, line 7: Awesome EditedTextHtmlParser class!

Tanvez (approved):
Looks good to me! Awesome work and your test really helped me understand how some of the parser is working!

lemmatized_text/views.py
```diff
@@ -81,6 +82,24 @@ def text(request, pk):
...
 
 
+def edit(request, pk):
+    lemmatized_text = get_object_or_404(models.LemmatizedText, pk=pk, created_by=request.user)
+
+    if request.method == "POST":
+        form = LemmatizedTextEditForm(request.POST)
+        if form.is_valid():
+            # pass to a "lemmatized text method" to handle changes
+            lemmatized_text.handle_edited_data(request.POST.get("title"), request.POST.get("text"))
```
arthurian, line 92: This could take a while, depending on how much of the text needs to be re-lemmatized, so we may want to turn `handle_edited_data` into an async job as a follow-up improvement.

lemmatized_text/models.py
```diff
@@ -198,6 +207,50 @@ def update_token(self, user, token_index, node_id, resolved):
...
+                f"{self.parse_following(token['following'])}</span>"
+            )
+            for token in self.data
+        ])
+
+    def parse_following(self, follower):
+        return follower.replace("\n", "<br/>")
+
+    def format_token_key_value_pairs(self, token):
+        return " ".join([f"{k}='{v}'" for k, v in token.items() if k not in ("word", "following")])
```
arthurian, line 252: Since `parse_following` and `format_token_key_value_pairs` appear to be pure helper functions for `transform_data_to_html`, you could consider nesting them inside instead of calling them as methods on the `LemmatizedText` model or extracting them as module-level functions. The only advantage is that it would make the dependency more explicit, and since the tests are only exercising `transform_data_to_html`, you can still be confident it's working as expected. It's also fine as-is, just thinking about how these are used compared to the other methods.

arthurian (approved):
Awesome work @dodget! The CKE + parser approach came together nicely. My only real concern before merging is the handling of the `createdAt` date, but otherwise I think it's good to go.

lemmatized_text/models.py
```diff
@@ -175,11 +183,12 @@ def api_data(self):
             "completed": self.completed,
             "tokenCount": self.token_count(),
             "lemmatizationStatus": self.lemmatization_status(),
-            "createdAt": self.created_at,
+            "createdAt": self.created_at.strftime("%b %d %Y, %I:%M%P"),
```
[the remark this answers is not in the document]
dodget: Ok, I'll fix that up

lemmatized_text/parsers.py
```diff
@@ -0,0 +1,122 @@
...
+        and sends new alpha numeric string to be lemmatized.
+        Returns None
+        """
+        followers = []
+        text = []
+        for idx, ch in enumerate(follower):
+            if ch.isalnum():
+                text = follower[idx:]
+                break
+            followers.append(ch)
```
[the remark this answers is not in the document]
dodget: Yes, exactly.

lemmatized_text/views.py
dodget: Agreed!

lemmatized_text/models.py
dodget: Good feedback, thanks

PR #329, opened 2021-07-28
Docker refactor

dodget:
This PR resolves inconsistencies in the Docker and Docker compose configuration of this project by:

- Consolidating to a single Dockerfile for local development and deployments (removing `Dockerfile.django`)
- Using a python base image and targeting the node version (14) to install
- Changing the docker compose `npm` service to use this project's Dockerfile
- Updating `.nvmrc` to version 14, to keep node versions consistent across local development options
- Updating the readme to include useful lessons learned when developing locally with docker-compose

## Notes
- If this PR is merged without updating some configurations in our ecs cluster, it is expected to break.
- One additional improvement to this PR would be to not set `USER root` in the Dockerfile. !Contribution welcomed!
- Once this PR is merged, and folks pull the changes locally, if they have been developing previously with Docker/docker-compose, they will likely need to remove their previous images/containers and re-run `docker-build`


dodget:
@bilbe 
@elliottyates (I thought you find this PR useful after our last discussion)

Dockerfile
```diff
@@ -1,15 +1,20 @@
...
 RUN npm i
 
-EXPOSE 80
+EXPOSE 8000 8080
 
-RUN npm rebuild node-sass
-RUN npm run build
\ No newline at end of file
+RUN npm rebuild node-sass && \
+    npm run build
```
arthurian, line 20: This works well as-is, but we might want to reorder the directives slightly so that the source code is added to the image after installing all the dependencies, but before doing a webpack build, that way any changes to the code don't automatically invalidate the other layers.

I did some experimenting, and this order seemed to work well:

```
FROM python:3.7
USER root
RUN mkdir -p /app
WORKDIR /app

# Install system dependencies
RUN curl -fsSL https://deb.nodesource.com/setup_14.x | bash - && \
    apt-get update && \
    apt-get install -y netcat git nodejs && \
    rm -rf /var/lib/apt/lists/*

# Install python dependencies
COPY ./Pipfile.lock ./Pipfile /app/
RUN pip install pipenv && \
    pipenv install --system --deploy && \
    rm Pipfile.lock Pipfile

# Install node dependencies
COPY ./package-lock.json ./package.json /app/
RUN npm i && \
    npm rebuild node-sass && \
    rm package-lock.json package.json  

# Copy app files
COPY . /app

# Build webpack assets 
RUN npm run build

# Use port 8000 for django server or 8080 for webpack dev server
EXPOSE 8000 8080
```

Note that `COPY` replaces `ADD` to copy the source code into the image. According to [docker best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy), `COPY` should be  preferred.

arthurian (approved):
Looks good @dodget! Tested locally after running `docker-compose down --rmi all` to remove all existing images and worked like a charm.

I made a small suggestion on the ordering of directives in the `Dockerfile` to make better use of the build cache. Also, I think this will actually work OK with our existing production deployment. After doing some more reading, the `EXPOSE` directive is considered documentation and doesn't actually do anything. And since we're running `gunicorn` and binding to port 80 in our docker entrypoint, I think it will work as-is. We'll have to do some testing to validate once this is merged, of course.

Tanvez (approved):
LGTM!

Dockerfile
```diff
@@ -1,15 +1,35 @@
...
+# Install system dependencies and target node v14
+RUN curl -fsSL https://deb.nodesource.com/setup_14.x | bash - && \
+    apt-get update && \
+    apt-get install -y netcat git nodejs && \
+    rm -rf /var/lib/apt/lists/*
+
+# Install python dependencies
+COPY ./Pipfile.lock ./Pipfile /app/
+RUN pip install pipenv && \
+    pipenv install --system --deploy && \
```
elliottyates, line 19: I'm still a little confused about what `--deploy` does. I see this:

> `--deploy` — Make sure the packages are properly locked in Pipfile.lock, and abort if the lock file is out-of-date.
https://pipenv-fork.readthedocs.io/en/latest/basics.html

> This will fail a build if the Pipfile.lock is out–of–date, instead of generating a new one. Or you can install packages exactly as specified in Pipfile.lock using the `sync` command.
https://pipenv-fork.readthedocs.io/en/latest/advanced.html#using-pipenv-for-deployments

docker-compose.yml
```diff
@@ -54,6 +57,8 @@ services:
             RQ_ASYNC: 1
             DATABASE_URL: postgres://hedera:hedera@postgres/hedera
             REDIS_URL: redis://redis:6379
+            PYTHONFAULTHANDLER: 1
```
elliottyates, line 60: Whoa, this looks pretty cool. What is this used for?

Dockerfile
```diff
@@ -1,15 +1,35 @@
...
+    rm Pipfile.lock Pipfile
+
+# Install node dependencies
+COPY ./package-lock.json ./package.json /app/
+RUN npm i && \
+    npm rebuild node-sass && \
```
elliottyates, line 25: Cute! I wonder why it still gave us an issue during the Aug 3 dojo.

Dockerfile
```diff
@@ -1,15 +1,35 @@
...
-RUN npm rebuild node-sass
-RUN npm run build
\ No newline at end of file
+
+# Create app working directory
+RUN mkdir -p /app
+WORKDIR /app
+
+# Install system dependencies and target node v14
+RUN curl -fsSL https://deb.nodesource.com/setup_14.x | bash - && \
```
elliottyates, line 11: I wonder if we could get it to use nvm and the `.nvmrc`.

Dockerfile
```diff
@@ -1,15 +1,35 @@
...
+COPY ./Pipfile.lock ./Pipfile /app/
+RUN pip install pipenv && \
+    pipenv install --system --deploy && \
+    rm Pipfile.lock Pipfile
+
+# Install node dependencies
+COPY ./package-lock.json ./package.json /app/
+RUN npm i && \
+    npm rebuild node-sass && \
+    rm package-lock.json package.json
```
elliottyates, line 26: I notice the package-lock.json is removed in this PR -- don't we want it there, to control dependencies?

elliottyates (commented):
Just some questions to help me understand the project and technologies better, thanks! :)

PR #332, opened 2021-08-03
Bookmark texts

### LINKED ISSUE #299: As a user, I want a reading bookmark on my dashboard so that I can return to what I was reading

arthurian:
This PR resolves #299.

- Added `LemmatizedTextBookmark` model to represent a user's text bookmarks.
- Added  `BookmarksListAPI` and `BookmarksDetailAPI` endpoints along with some unit tests in `hedera.tests`.
- Updated `App.vue` and `Learner.vue` so that a user can bookmark the text on the teacher and learner views of the lemmatized text.
- Updated the `Dashboard.vue` to display a user's bookmarks. The bookmarks on the dashboard are displayed in the order in which they were added.
- Added `hedera/tests/utils.py` with some helper utilities for unit tests.

Notes:
- Still need to add some Vue frontend tests and some backend failure and edge case tests.


hedera/tests/utils.py
```diff
@@ -0,0 +1,47 @@
+from django.contrib.auth import get_user_model
+
+from lemmatized_text.models import LemmatizedText, LemmatizedTextBookmark
+
+
+User = get_user_model()
+
+
+def create_user(**kwargs):
```
Tanvez, line 9: awesome utils file! Will keep this in mind for future testing

lemmatized_text/models.py
```diff
@@ -214,3 +224,31 @@ def data(self):
...
         )
+
+
+class LemmatizedTextBookmark(models.Model):
+    user = models.ForeignKey(
+        getattr(settings, "AUTH_USER_MODEL", "auth.User"),
+        on_delete=models.CASCADE
+    )
+    text = models.ForeignKey(LemmatizedText, on_delete=models.CASCADE)
+    created_at = models.DateTimeField(auto_now_add=True)
```
dodget, line 235: Since this is a 1:1 relationship I was wondering why a `bookmark` boolean field couldn't just be added to the `LemmatizedText` model. But it's nice to have the `created_at` field, and having its own model may allow us to add things like the bookmark's position in the text later if we want to.
Are those things you considered when adding a new model? 👍 Just kibitzing

Tanvez (approved):
LGTM!

static/src/js/app/Dashboard.vue
```diff
@@ -9,14 +9,16 @@
     <button type="button" class="btn btn-primary" @click="openModal">
       Quick Add Vocab
     </button>
+    <BookmarkList></BookmarkList>
```
dodget, line 12: Can this element be self-closing?
https://vuejs.org/v2/style-guide/#Self-closing-components-strongly-recommended

dodget (approved):
Great work Artie!

static/src/js/app/Dashboard.vue
arthurian: Absolutely! I didn't know about self-closing components, so thanks for the pointer. I'll fix that.

lemmatized_text/models.py
arthurian: Great question. Since users can bookmark multiple texts, and texts can be bookmarked by multiple users, I modeled it as a many-to-many relationship (`Users ↔ LemmatizedText`). I thought about adding a `ManyToManyField` to the `LemmatizedText` model, but decided against it because I also wanted to record additional information such as the date the user bookmarked the text. And like you said, we might also want to record the bookmark's position in the text, so by making that intermediate model explicit, we can add those enhancements relatively easily.

PR #333, opened 2021-08-03
Personal vocab delete

Tanvez:
Related to [ATGU-3162](https://jira.huit.harvard.edu/browse/ATGU-3162)
- Added delete button for personal vocab list
- Added delete route to Django
- Added test for vue and Django
- Fixed Mobile view for delete button - button trails off the screen for mobile view


hedera/api.py
```diff
@@ -297,6 +301,15 @@ def post(self, request, *args, **kwargs):
...
         return JsonResponse({"data": vl.data()})
 
+    def delete(self, request, *args, **kwargs):
+        data = json.loads(request.body)
+        if "id" in data:
+            count, _ = PersonalVocabularyListEntry.objects.filter(id=data["id"]).delete()
+            if(count != 0):
+                return JsonResponse({"data": True, "id": data["id"]})
+            return HttpResponseNotFound(f"could not find vocab with id={data['id']}")
+        return HttpResponseBadRequest(f"missing 'id'")
```
dodget, line 311: No need for an f string here, if there is no expression/variable to evaluate.

hedera/tests/test_api.py
```diff
@@ -25,3 +38,21 @@ def test_post_personal_vocabulary_quick_add_api(self):
...
+    def test_unsuccessful_delete_personal_vocab_entry_empty_request_body(self):
+        response = self.client.delete("/api/v1/personal_vocab_list/", json.dumps({}), content_type="application/json")
+        self.assertEqual(response.status_code, 400)
+
+    def test_unsuccessful_delete_personal_vocab_entry_id_does_not_exist(self):
+        payload = {
+            "id": 9999
+        }
+        response = self.client.delete("/api/v1/personal_vocab_list/", json.dumps(payload), content_type="application/json")
+        self.assertEqual(response.status_code, 404)
```
dodget, line 58: Love it.

static/src/js/tests/PersonalVocab.spec.js
```diff
@@ -0,0 +1,55 @@
+import { shallowMount, createLocalVue, mount } from '@vue/test-utils';
```
dodget, line 1: Nice to know that component mounting functions are `@vue/test-utils`

dodget (approved):
Nice work 👍

hedera/api.py
Tanvez: ahh should i update it to a json response with the error message as payload?

arthurian (approved):
Looks good @Tanvez! :+1:

The only thing that might be worth revisiting is error responses to the client. Since this is a cross-cutting concern between this and some other outstanding PRs, it may be worth merging the existing approved PRs and following up on that separately.

PR #337, opened 2021-09-07
Pin python dependencies

arthurian:
This PR pins the python dependencies to avoid unintentional drift, such as the drift from Django 2.x to 3.x.

**Notes:**
- The main packages have been pinned to the exact version that was most recently deployed, and the dev packages are constrained to minor version upgrades.
- `Django` is now pinned to v3.2.5.
- There are many packages with newer versions, so we will want to follow-up with updates to any packages that make sense. For example, `boto3` is now on v1.18.36 and we are running v1.12.43.


dodget (approved):
Looks good, thanks Artie

Tanvez (approved):
LGTM!

arthurian:
Per follow-up discussion this morning, we're going to relax package requirements so that they get patch updates (~=).

PR #342, opened 2021-09-13
Fix personal vocab perfomance issue

Tanvez:
Updated PersonalVocabularyListEntry model to use node_id field

Total queries have been reduced from `N`(number of entries in the `vocab_list_personalvocabularylistentry` table ) calls to `lattices_latticenode`(to retrieve the pk) to 4 calls shown below

The following are the calls to their respective tables:
  - auth_user
  - vocab_list_personalvocablist
  - vocab_list_personalvocabularystats,
  - vocab_list_personalvocabularylistentry 
```
hedera-django    | http://0.0.0.0:8000/api/v1/personal_vocab_list/
hedera-django    | |------|-----------|----------|----------|----------|------------|
hedera-django    | | Type | Database  |   Reads  |  Writes  |  Totals  | Duplicates |
hedera-django    | |------|-----------|----------|----------|----------|------------|
hedera-django    | | RESP |  default  |    4     |    0     |    4     |     1      |
hedera-django    | |------|-----------|----------|----------|----------|------------|
hedera-django    | Total queries: 4 in 0.0858s 

```


vocab_list/models.py
```diff
@@ -283,7 +283,7 @@ def data(self):
             headword=self.headword,
             gloss=self.gloss,
             familiarity=self.familiarity,
-            node=self.node.pk if self.node is not None else None,
+            node=self.node_id if self.node_id is not None else None,
```
dodget, line 286: Woah, did this take care of it? That's awesome.

Tanvez: yea it did! Still unsure about why there was calls to the lattice node table if the node_id was already present on the object.

dodget: Its not totally clear to me either but maybe by calling `self.node.pk` you are instantiating a node object and then getting the pk, where calling `self.node_id` is using the id that is used for the foreign key relationship. So in the latter, there is no query required. I've convinced myself actually.

dodget (approved):
Very simple, very awesome 👍

arthurian (approved):
Nice job resolving the django query issue @Tanvez. The API query is now returning very quickly, although I noticed in my own testing that a personal vocab list of 813 items takes ~50 seconds to fully render and my browser freezes for a good chunk of that time. It looks like there may be a Vue issue at play as well, although I'm not certain what specifically is causing the poor performance.  I think we'll need a follow-up story to either implement pagination or fix whatever is causing the slow rendering (assuming it's not just me).

Tanvez:
> Nice job resolving the django query issue @Tanvez. The API query is now returning very quickly, although I noticed in my own testing that a personal vocab list of 813 items still takes ~50 seconds to fully render and my browser freezes for a good chunk of that time. It looks like there may be a Vue issue at play as well, although I'm not certain what specifically is causing the poor performance. I think we'll need a follow-up story to either implement pagination or fix whatever is causing the slow rendering (assuming it's not just me).

Thanks Artie! There is still noticeable delay for me as well. Definitely agree with you on a follow-up story to implement a fix with either pagination or find the cause of the slow rendering.

PR #343, opened 2021-09-14
Paginate personal vocab entries

arthurian:
This PR fixes slow render performance of the personal vocab when there are a lot of entries.

Changes:
- Added [vue-good-table](https://xaksis.github.io/vue-good-table/) component to render the personal vocab entries.
- Enabled [pagination](https://xaksis.github.io/vue-good-table/guide/configuration/pagination-options.html#enabled) and [searching](https://xaksis.github.io/vue-good-table/guide/configuration/search-options.html) of the entries. 

Notes:
- Since the API query returns quickly and is not currently a bottleneck, we're not paginating that yet. The `vue-good-table` component supports server-side pagination, so we can implement that when needed.

Example screenshot:

hedera-vue-good-table-20210914

### DIFF
--- a/static/src/js/app/index.js
+++ b/static/src/js/app/index.js
@@ -1,6 +1,8 @@
 import Vue from 'vue';
 import VueKeybindings from 'vue-keybindings';
 import VueFilterDateFormat from '@vuejs-community/vue-filter-date-format';
+import VueGoodTablePlugin from 'vue-good-table';
+import 'vue-good-table/dist/vue-good-table.css';
 
 import globalComponents from './components';
 import store from './store';
@@ -30,6 +32,7 @@ const load = (appId, appComponent, keyBindingsConfig, appProps) => {
     }
 
     Vue.use(VueFilterDateFormat);
+    Vue.use(VueGoodTablePlugin);
 
     /* eslint-disable no-new */
     new Vue({
--- a/package-lock.json
+++ b/package-lock.json
@@ -1947,7 +1947,7 @@
         "source-map": {
           "version": "0.6.1",
           "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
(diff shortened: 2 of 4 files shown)


static/src/js/app/index.js
```diff
@@ -1,6 +1,8 @@
 import Vue from 'vue';
 import VueKeybindings from 'vue-keybindings';
 import VueFilterDateFormat from '@vuejs-community/vue-filter-date-format';
+import VueGoodTablePlugin from 'vue-good-table';
+import 'vue-good-table/dist/vue-good-table.css';
```
dodget, line 5: should this be imported in the css section of `PersonalVocab.vue`? Not sure what is most idiomatic.

dodget (approved):
This is great. Did you experiment with `dropdownAllowAll: true`? I guess allowing up to 100, plus search functionality is just fine!

Tanvez (approved):
Looks good to me! Awesome work Artie

static/src/js/app/index.js
arthurian: I'm not entirely sure. Since the `vue-good-table` component can be used anywhere, it seemed like it should be a global style instead of scoped to `PersonalVocab.vue`. But if it's only going to be used in that one component, then we should probably move it there. For now, I'm leaning toward keeping the two imports together in the `index.js`.

arthurian:
> This is great. Did you experiment with `dropdownAllowAll: true`? I guess allowing up to 100, plus search functionality is just fine!

I tried it with the `All` option in the drop-down, and it's still pretty slow, but I suppose there's no harm in adding that option as long it's not the default. If the user only has a few hundred entries, it will probably be fine.

static/src/js/app/index.js
dodget: 👍

PR #344, opened 2021-09-16
added supported languages endpoint

Tanvez:
- Added supported languages endpoint
- Added test

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
         - uses: actions/checkout@v2
 
         - name: Setup Python
-          uses: actions/setup-python@v1
+          uses: actions/setup-python@v2
           with:
             python-version: 3.7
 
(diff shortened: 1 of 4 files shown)


arthurian (approved):
Looks good! 👍

dodget (approved):
👍

.github/workflows/ci.yaml
```diff
@@ -26,18 +26,14 @@ jobs:
         - name: Setup flake8 annotations
           uses: rbialon/flake8-annotations@v1
 
-        - name: Install Python Dependences
-          if: steps.pip-cache.outputs.cache-hit == false
-          run: pipenv install --dev
+        - name: Install Python Dependences, run linting and import sorting
```
elliottyates, line 29: Just curious @Tanvez -- what's the rationale for putting these all in one step? I'm not very familiar with github actions. Thanks!

Tanvez: I think we wanted it to be more concise. The downside of combining the commands is if any of the commands fail it would be hard to figure out which one due to the nature of running the commands in one scripts. Maybe @dodget can help explain it better than I can.

dodget: This was of my creation, so I'll pipe in.
We were getting an error starting in the `isort` step, saying that pipenv was using python 3.8 although the pipfile requires python 3.7. For whatever reason, I think that each step was creating its own pipenv virtual environment, and not using the environment created in previous steps. So I threw all the python related checks into one step, and we are back in business.
I think previously, the workflow was in a state of working, but also being broken. I think we were just lucky that `ubuntu-latest` had a system python of 3.7. Now that system python appears to be 3.8, and the broken parts revealed themselves.

PR #345, opened 2021-09-17
Add personal vocab entry to store immediately

dodget:
This PR resolves [ATGU-3234](https://jira.huit.harvard.edu/browse/ATGU-3234) by:

- Adjusting the `CREATE_PERSONAL_VOCAB_ENTRY` mutation to add the returned data (the entry that has been added) to the front of `state.personalVocabList.entries`. This will add the entry to the top of the table, unsorted.
- Adding tests to validate the functionality

## Notes

- I adjusted the mutation tests to use the canonical state, rather than each unit test stubbing out a mock state
- I changed the initial state's `personalVocabList` from `null` to `{}`. If we know its going to be an object, what initially set it as null? This seems more explicit, and may save some unwanted errors down the line
- There are a ton of commits here, mainly attempting to get the ci to work. There has to be a cleaner way to iterate on that


static/src/js/app/vuex/mutations.js
```diff
@@ -68,6 +68,9 @@ export default {
   },
   [CREATE_PERSONAL_VOCAB_ENTRY]: (state, data) => {
     state.personalVocabAdded = data.created;
+    if (state.personalVocabList.entries) {
+      state.personalVocabList.entries = [data.data, ...state.personalVocabList.entries];
```
elliottyates, line 72: I can understand how this allows someone to see the new entry right away, even in a long list, but would it match the actual order? It might be confusing to refresh the screen and see that item disappear (in an unsorted table). On that note, would a data table with an existing sort or filter automatically refresh its pages and sort order (as that might hide the entry as well)?

static/src/js/app/vuex/tests/mutations.spec.js
```diff
@@ -23,21 +21,49 @@ describe('Mutations', () => {
...
+      gloss: 'bar',
+      familiarity: 1,
+      node: null,
+    };
+    const data = {
+      created: true,
+      data: vocabData,
+    };
+
+    it('successfully updates the state without entries fetched', () => {
```
elliottyates, line 36: I don't really understand what this is testing -- the opposite of the one below? Why?

package-lock.json
```diff
@@ -1947,7 +1947,7 @@
         "source-map": {
           "version": "0.6.1",
           "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
-          "integrity": "sha512-UjgapumWlbMhkBgzT7Ykc5YXUT46F0iKu8SGXq0bcwP5dz/h0Plj6enJqjz1Zbq2l5WaqYnrVbwWOWMyF3F47g==",
+          "integrity": "sha1-dHIq8y6WFOnCh6jQu95IteLxomM=",
```
elliottyates, line 1950: Seems like the `package-lock.json` wouldn't need to change for this update?...

arthurian (approved):
Looks good @dodget! 👍 

_Re: github actions_ - while I didn't have a ton of success using [act](https://github.com/nektos/act) locally a few months ago, I admittedly didn't spend a ton of time trying to get it to work. Perhaps it's worth us collectively investing some time into that tool to see if we can make that workflow less painful.

elliottyates (commented):
Just some thoughts! Also there is chatter on the netz about https://github.com/nektos/act as a way to test CI locally. You can also, of course, do a `rebase -i` to make the commit history cleaner (github is pretty savvy about it, too, in terms of connecting comments to rebased work).

static/src/js/app/vuex/mutations.js
dodget: It does not respect any ordering. I agree that it could be confusing in the way that you've described, but the other implementations that I thought of had equally confusing UX. I think its a 6 in one, half dozen in another situation. This seemed like a fine enough solution until there is some actual feedback after folks have been using the new table.

static/src/js/app/vuex/tests/mutations.spec.js
dodget: Tests what would happen if you use the quick add vocab feature if your personal vocab list data has not yet been fetched. For instance, a user could use the quick add vocab button, but not have visited the personal vocab page, which would mean that the users personal vocab entries have not been fetched, and therefor `.entries` should still be `undefined`.

The one below is the same test, but where the entries have been fetched, and it tests that `.entries` is mutated.

package-lock.json
dodget: Seems like it, but I always commit changes to lockfiles. I assume this was the result of rebuilding the docker image, but I could remove it

Tanvez (approved):
LGTM!

PR #348, opened 2021-09-22
Manage vocab text

dodget:
This PR is the first step in resolving [ATGU-2912](https://jira.huit.harvard.edu/browse/ATGU-2912) by:

- Adding a management command to delete all data from lattices, lemmatized_text, and vocab_list
- Adding documentation to the README on how to execute a "resetting" of the DB

## Notes

- This PR also remove the `databasetext` app that is unused cruft. To my knowledge and testing, this has no effect on the project


hedera/management/commands/reset_db.py
```diff
@@ -0,0 +1,60 @@
...
+            if not no_input:
+                confirmation = input(f"Delete all objects from {model}? y/n ")
+                if confirmation == "y":
+                    self.exectute_deletion(model)
+                else:
+                    self.stdout.write(f"Skipping {model} data deletion")
+            else:
+                self.exectute_deletion(model)
+
+    def exectute_deletion(self, model):
```
arthurian, line 58: typo: `execute_deletion `

.pre-commit-config.yaml
```diff
@@ -1,3 +1,5 @@
+default_language_version:
+  python: python3.7
```
arthurian, line 2: Could this be set more generically to `python3` instead of `python3.7`, or does it need to be explicit?

README.md
```diff
@@ -161,3 +161,21 @@ pre-commit install
 ```
 You're all set! Now when you commit code upstream the commit will trigger the different linters and flag the required changes
 Notes: you can quickly fix javascript eslint errors with `npm run lint:fix` 
+
+
+## Resetting the Database
+
+Every so often, it may be requested that the dev DB be "reset." Meaning, certain tables be truncated, and the data be reloaded. You can do that with the following set of commands. Note that these commands are for executing against a running docker container. If you are running the project in another way, adjust the incantations accordingly.
```
arthurian, line 168: It might be worth mentioning that this preserves user accounts, so any users who have signed up will still be able to login after the reset.

arthurian (approved):
Looks good! AFAIK the `databasetext` app is not used at all (was part of an early version to capture the raw texts). :+1:

.pre-commit-config.yaml
dodget: I'm fairly sure it could be `python3`, but since we are using 3.7 on this project, I thought I'd set it explicitly and maintain consistency. I'd be happy to change it if you have a strong feeling about it.

hedera/management/commands/reset_db.py
dodget: Thanks

README.md
dodget: Good idea. I've made more explicit.

hedera/management/commands/reset_db.py
```diff
@@ -0,0 +1,60 @@
...
+    LemmaNode,
+    LemmatizationLog,
+    LemmatizedTextBookmark,
+    VocabularyListEntry,
+    PersonalVocabularyListEntry,
+    PersonalVocabularyStats
+]
+
+
+class Command(BaseCommand):
```
Tanvez, line 33: Awesome! Will refer to this in the future for any situation calling for management commands!

Tanvez (approved):
Looks good to me!

PR #349, opened 2021-09-22
Chinese text support MVP

### LINKED ISSUE #304: As a student I want to be able to mark words from a Chinese text to my personal word list

### LINKED ISSUE #305: As a student, I want to be able to read a lemmatized Chinese text

### LINKED ISSUE #307: As a teacher, I want to be able to import a Chinese text and see a lemmatized version in the app.

elliottyates:
# Purpose

Adds Chinese text support to Hedera. Supports simplified and traditional characters (separately or mixed). It consists of:

- Chinese tokenizer: [LAC](https://github.com/baidu/lac) _("Lexical Analysis of Chinese")_
    - uses pypi's v2.1.2, which [isn't tagged in github](https://github.com/baidu/lac/compare/v2.1.0...master)
    - License: [Apache 2.0](https://github.com/baidu/lac/blob/3e10dbed9bfd87bea927c84a6627a167c17b5617/LICENSE) _(this commit represents the pypi v2.1.2 release)_
- Chinese-context lattice: [CC-CEDICT dictionary](https://cc-cedict.org/wiki/)
    - License: [Creative Commons Attribution-ShareAlike 4.0 International License]((https://www.mdbg.net/chinese/dictionary?page=cedict))

Related stories:
- Closes #307 (instructor can import and lemmatize a Chinese text)
- Closes #305 (student can view a lemmatized text).
- Closes #304 (student can compare text with personal vocab list)

# Using features

There's now a "Chinese" language listed in the form for importing a new text. It can be used to parse a text and look for existing lemmas in the lattice.

Parsing a text will show lemmas properly separated, but will not show any glosses unless some are loaded into the lattice. That's not trivial, so see below under the heading "Loader performance" for details (there is a test dict that can be used locally to play with the feature).

If you do load the non-trivial test data (`cedict_ts.u8`) into the lattice, you can then import the following sample text to see the various functions:

```
1. 該爆料最初由《華盛頓郵報》資深記者鮑勃·伍德沃德（Bob Woodward）和羅伯特·科斯塔（Robert Costa）在即將於下周出版的新書《危險》（Peril）中披露。美聯社援引一名高級國防官員證實了書中說法。

2. 该爆料最初由《华盛顿邮报》资深记者鲍勃·伍德沃德（Bob Woodward）和罗伯特·科斯塔（Robert Costa）在即将于下周出版的新书《危险》（Peril）中披露。美联社援引一名高级国防官员证实了书中说法。

3. 你好我是個｜人, 你好我是個|人, 你好我是個人

4. 哈＿哈＿哈, 哈_哈_哈, 哈哈哈, 哈哈|哈
```

Note the following:

- Line 1: **Punctuation, non-Chinese text** The parser handles Chinese punctuation, Latin characters, the "middle dot" used in Chinese names, etc. in a reasonable way.
- Lines 1 and 2: **Multiple forms** - Line 1 is traditional, and line 2 is simplified; they are semantically identical. Note that selecting a word whose form is the same in both traditional and simplified, such as `中`, will highlight both occurrences, while selecting a word whose form is different, such as `書` (simplified equivalent: `书`), does not highlight the occurrence with a different form.
- Lines 3 and 4: **Full-width characters** - the first block of text (before the comma) shows full-width pipes and underscores, and the second block uses latin characters. They are both handled correctly as "control signals" by the parser.
- Line 4: **Ambiguous lemma resolution** - The last block `哈哈|哈` is parsed into `哈哈` and `哈` -- the latter, individual character shows the ability to resolve ambiguous lemmas (there are multiple dictionary headwords for this character).

# Automated testing

Some new tests were added to:
- test that LAC tokenizes as expected
- test that our lemmatizer correctly outputs "triples" that can be used for creating `LemmaNode`s
- check that the lemmatizer/tokenizer handles the pipe (`|`) and underscore (`_`) characters correctly (using them to split a raw token into multiple lemmas and to join multiple raw tokens into a single lemma, respectively).

```
docker compose run django python manage.py test lemmatization.tests
```

# Details

## Simplified and traditional forms

Chinese comes in two main textual representations: [simplified](https://en.wikipedia.org/wiki/Simplified_Chinese_characters) and [traditional](https://en.wikipedia.org/wiki/Traditional_Chinese_characters). The approach taken here is to treat the two forms as a single language, per ISO 639.2 (both are represented by `zho`), and allow a Chinese text to contain either or both forms. The lattice thus needs a headword for each form; for the moment these are duplicates (they only differ by headword). This means that glosses are the same for each "form" of a Chinese word, but when highlighting words in the text the different forms are not unified (highlighted together), as they are in fact different lemmas (note that changing this behavior may require changing the underlying data model or at least the way lemma are associated, so that the text lemma highlighting function can match different forms together).

Here's an example of CEDICT entries (/traditional form/simplified form/pronunciation/glosses) that are loaded into the lattice (see the readme for full lattice loading instructions, but first read the "Loader performance" section below). Note that the first entry has form differences, and the second does not.

```
一呼百應 一呼百应 [yi1 hu1 bai3 ying4] /a hundred answers to a single call (idiom)/to respond en masse/
一味 一味 [yi1 wei4] /blindly/invariably/
```

Lattice definitions include the simplified and traditional form, mandarin pronunication, and `/`-delimited glosses, all directly from CEDICT. The forms are included so that a reader can reference and be aware of both forms for any lemmas that have different forms. This is was an expediency to avoid having to change the data model; pedagogical needs might indicate a need for separate fields (e.g. so the reader won't see "forms" as part of a gloss, for review purposes), or e.g. for showing the two forms in the gloss only if they differ.

I also noticed that the current font represents Chinese poorly; traditional and simplified characters display a bit differently. Here's a screenshot:

[
Screen Shot 2021-09-21 at 9 57 49 PM
](url)

## Loader performance

The full dictionary ("chinese"-context lattice) is 120k+ entries; also, many entries need to be effectively duplicated (when the traditional and simplified forms of the lemma are different, they are saved as separate lattice nodes so they can be looked up using either form). This results in a substantially long single-threaded sequential db load time, as I simply cribbed from the approach from the clancy (Russian) loader.

This could be ameliorated by making use of workers (asynchronous loading), and chunking the input into jobs (concurrent loading); it could also be fixed by making the `ChineseService` an API for accessing the dictionary via a database adapter (e.g. using SQLite), and load the lattice nodes only upon first lookup in the DB, similar to what the `MorpheusService` does.

For the moment, another way to test dict entries in a local app without loading the full set of entries is to create a smaller version of the `cedict_ts.u8` file and load it, instead, ensuring it has entries for the text you wish to test. For example, this short text "你好我是個｜人" should only need the following lines in the dictionary file (this file `cedict_ts_sampled_for_testing.u8` is provided in the `import-data` dir):

```
# here's an example sentence that will show some of these glosses once parsed:
#   你好我是個｜人
# note that the parser should separate the 個｜人 into two lemmas, such that
# they will will be glossed with 個 and 人, instead of the compound word 個人,
# which would be the natural token parsed by the tokenizer if the pipe char |
# were not included.
你好 你好 [ni3 hao3] /hello/hi/
我 我 [wo3] /I/me/my/
是 是 [shi4] /is/are/am/yes/to be/
個 个 [ge4] /individual/this/that/size/classifier for people or objects in general/
個人 个人 [ge4 ren2] /individual/personal/oneself/
人 人 [ren2] /person/people/CL:個|个[ge4],位[wei4]/
```

To load that small set of data, simply update line 10 of `load_chinese_lattice.py` to point to the smaller set of test data:

```python
DICT_FILE = "import-data/cedict_ts_sampled_for_testing.u8"
```

Then run:

```sh
docker-compose run django python manage.py shell -c "import load_chinese_lattice"
```

## Next steps

- [x] fix seemingly unrelated failing test `FAIL: test_handle_edited_data (lemmatized_text.tests.LemmatizedTextTests)` ([fixed](https://github.com/Hedera-Lang-Learn/hedera/pull/349#issuecomment-925180240) in #348)
- [ ] load chinese data quicker (e.g. in separate data source, or in parallel into lattice)
- [ ] select and test a better font for Chinese (simplified/traditional)
- [ ] deploy to a test environment for wider review


Pipfile
```diff
@@ -26,8 +26,10 @@ sentry-sdk = "~=1.3.1"
 rq = "~=1.9.0"
 wagtail = "~=2.14.1"
 django-lti-provider = "~=0.4.4"
+lac = "==2.1.2"
 
 [dev-packages]
+debugpy = "~=1.4.3"
```
elliottyates, line 32: I wanted to add watchmedo via the [`watchdog` package](https://github.com/gorakhargosh/watchdog) to manage RQ worker reloading, but apparently [pipenv can't handle `extras` (optional dependencies) properly via the `Pipfile.lock`](https://github.com/pypa/pipenv/issues/4407) just yet.

Pipfile
```diff
@@ -26,8 +26,10 @@ sentry-sdk = "~=1.3.1"
 rq = "~=1.9.0"
 wagtail = "~=2.14.1"
 django-lti-provider = "~=0.4.4"
```
elliottyates, line 28: I tried to suppress `RemovedInDjango40Warning` log warnings from `lti_provider` module, but this is a bit of a pain as it appears that we'd need to do some modifications to the log config, and as there is only a single `settings.py` right now it seemed too intrusive. It can also be done by setting `DEBUG = False` in settings, but that has larger effects than simply cleaning up log output.

hedera/api.py
```diff
@@ -135,14 +135,15 @@ def get_data(self, new_text=None):
         )
 
     def post(self, request, *args, **kwargs):
+        cloned = None
```
elliottyates, line 138: Fixes issue `"cloned" is possibly unbound.`

load_chinese_lattice.py
```diff
@@ -0,0 +1,108 @@
...
+from typing import Optional, Tuple
+
+from lattices.models import LatticeNode, LemmaNode
+
+
+DICT_FILE = "import-data/cedict_ts.u8"
+DEFAULT_LEMMA_NODE_CONTEXT = "chinese"
+
+
+def create_lemma_node(lemma, lattice_node, context=DEFAULT_LEMMA_NODE_CONTEXT):
```
elliottyates, line 14: This section is exactly the same as the clancy loader.

load_chinese_lattice.py
```diff
@@ -0,0 +1,108 @@
...
+
+ROW_COUNT_FOR_UPDATE = 100
+ROW_TO_START_FROM = 98500
+current_row = 0
+start_time = time.time()
+block_start_time = start_time
+rows = []
+
+with open(DICT_FILE) as f:
+    rows = f.readlines()
```
elliottyates, line 62: This was to see if loading everything into memory would improve performance -- it wasn't the bottleneck, though.

docker-compose.yml
```diff
@@ -75,6 +80,9 @@ services:
             RQ_ASYNC: 1
             DATABASE_URL: postgres://hedera:hedera@postgres/hedera
             REDIS_URL: redis://redis:6379
+            PYTHONDONTWRITEBYTECODE: 1
+            PYTHONUNBUFFERED: 1
         networks:
             - localdev
-        command: /bin/sh -c 'while ! nc -z postgres 5432; do sleep 1; done; while ! nc -z redis 6379; do sleep 1; done; python manage.py rqworker ${RQ_QUEUES:-default}'
+        command: /bin/sh -c 'while ! nc -z postgres 5432; do sleep 1; done; while ! nc -z redis 6379; do sleep 1; done; exec python manage.py rqworker ${RQ_QUEUES:-default}'
+        stop_signal: SIGINT
```
elliottyates, line 88: This speeds up recycling containers. The default stop signal Docker sends to the container is SIGTERM, which Django never receives, so Docker waits ten seconds before killing the container with SIGKILL. Changing the stop signal to SIGINT will cause Docker to send Ctrl-C to the container; using `exec` in the script will cause that (last) script command to assume PID 1, and thereby allow Ctrl-C to be propagated to Django, resulting in a very fast graceful shutdown (usually less than a second).

This is particularly important for the `worker` container, as it doesn't auto-reload worker code.

docker-compose.yml
```diff
@@ -59,13 +60,17 @@ services:
             REDIS_URL: redis://redis:6379
             PYTHONFAULTHANDLER: 1
             PYTHONUNBUFFERED: 1
+            PYTHONDONTWRITEBYTECODE: 1
```
elliottyates, line 63: This ensures that pyc files are not being written in development, so that changes to py files on the host are immediately represented in the running container.

load_chinese_lattice.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def node_exists(lemma: str) -> bool:
+    return LemmaNode.objects.filter(lemma=lemma, context=DEFAULT_LEMMA_NODE_CONTEXT).exists()
+
+
+row_grouping_re = re.compile(r"^(?P<lemma>\S+) (?P<altform>\S+) (?P<defn>.*)$")
+
+ROW_COUNT_FOR_UPDATE = 100
+ROW_TO_START_FROM = 98500
```
elliottyates, line 55: typo - this needs to be `1`.

elliottyates:
> I have a fix for the `test_handle_edited_data` in my current PR #348
> That test is a bit wonky, but I haven't thought of a good way to make it more resilient. I'll try my hand at it next time.

Thanks for looking into that! (Working through tests could be a good way for the rest of us to get to know the feature code better too -- maybe it can be a dojo topic one day.)

hedera/templates/lemmatized_text/create.html
```diff
@@ -19,6 +19,7 @@ <h1>New Text to Lemmatize</h1>
     <div class="form-group">
       <label>Language</label>
       <select name="lang" class="form-control">
+        <option value="zho" {% if select_lang == "zho" %}selected{% endif %}>Chinese</option>
         <option value="grc" {% if select_lang == "grc" %}selected{% endif %}>Greek</option>
         <option value="lat" {% if select_lang == "lat" %}selected{% endif %}>Latin</option>
         <option value="rus" {% if select_lang == "rus" %}selected{% endif %}>Russian</option>
```
dodget, line 25: Not part of the scope of this PR, but this should be refactored to loop over the supported languages defined in settings.

lemmatization/services/chinese.py
```diff
@@ -0,0 +1,223 @@
...
+        # that lattice lookups will be optimal
+        tokens = self.re_tokenize(tokens)
+
+        # 5. Generate stream of lemmatized tokens
+        lemma_triples = get_triples(tokens)
+
+        return lemma_triples
+
+    def prepare_for_segmentation(self, text: str) -> str:
+        """ Override this method """
```
dodget, line 59: Does this docstring imply that these methods must be overridden in order for things to work properly? (not sure if that sentence makes sense)

If so, what do you think about `raise NotImplementedError` to reinforce that idea. [Link to docs](https://docs.python.org/3/library/exceptions.html#NotImplementedError)

lemmatization/services/chinese.py
```diff
@@ -0,0 +1,223 @@
...
+    normalized_text = unicodedata.normalize("NFC", text)  # normalize
+    return normalized_text.replace("\uFF5C", "|").replace("\uFF3F", "_")
+
+
+def normalize_token(token: str) -> str:
+    """
+    Perform transformations or adjustments to individual Chinese tokens
+    """
+    # currently a no-op
+    return token
```
dodget, line 146: What do you consider the use of defining a function that only returns its argument? Just curious.

dodget (approved):
Looks great @elliottyates ! The tests are super clear and provide so much confidence. I just added a couple of questions/suggestions

lemmatization/services/chinese.py
elliottyates: Hmm, yes, a good question. I defined this function to provide a place for any future normalization to happen, and as an indication that the second part of the "triple" is meant to be a normalized form (is it documented anywhere?). Also the process evolved by following the pattern of the other lemmatizers. But now that I look at it, there seem to be two approaches that would be better: it could simply be gone to reduce confusion, and returning `(token, token, following)` can be documented to support that choice; or, if we left it, it makes more sense for it to be a class-specific method, as it would likely vary with any given implementation. 

In the end when looking at it I opted with c791be7 to make everything encapsulated, either in the base class or the LAC subclass. I hope it's easier to understand that way, and supports easier alternative subclassing, but it might be overkill in some people's eyes.

lemmatization/services/chinese.py
elliottyates: A better idea for sure! Thanks! Done in c791be7.

load_chinese_lattice.py
elliottyates: Addressed in 4350339.

Tanvez:
Awesome work Elliott! I tested the lemmatization and from what I can see it matches the segmentation found in your test files. I tried to test out some single characters to see if they would get the correct lemma but some of the characters like `你` and a random phrase `据我看` (taken from the cedic_ts file) doesn't get linked to a lemma.
Maybe I did not import the `cedict_ts.u8` correctly. Do I need to download a language package for MacOS? I am seeing some entries in the database look like this:
_**lattice_latticenode Table**_
Screen Shot 2021-09-27 at 3 07 25 PM

_**lattice_lemmanode Table**_
Screen Shot 2021-09-27 at 3 06 20 PM

FYI: When I first ran the import for cedict_ts.u8 it "`Processed 120182 rows in 547.9292409420013 seconds`" on a second run to make sure I wasn't missing any data it took about `143.27139854431152 seconds`

elliottyates:
> I tried to test out some single characters to see if they would get the correct lemma but some of the characters like `你` and a random phrase `据我看` (taken from the cedic_ts file) doesn't get linked to a lemma. Maybe I did not import the `cedict_ts.u8` correctly. 

Huh, that's strange. Here's a [screen capture of what I get when I test those phrases](https://user-images.githubusercontent.com/8975317/134972590-00b97a12-3d1a-4159-a660-4ce200800535.mov); basically, the LAC segmenter (tokenizer/lemmatizer) is not considering "据我看" to be a single token, but rather that "phrase's" constituents are separate tokens. There's not much we can do to match the dictionary perfectly to the segmenter other than train our own models or switch up the segmenter.

> FYI: When I first ran the import for cedict_ts.u8 it "`Processed 120182 rows in 547.9292409420013 seconds`" on a second run to make sure I wasn't missing any data it took about `143.27139854431152 seconds`

This looks very, very fast to me -- it took many hours on my laptop. I think that you likely ran the load script before I fixed it in 4350339 -- if you try it with that `ROW_TO_START_FROM` set to `1` you'll be loading the whole shebang... but only if you want to. ;)

> Do I need to download a language package for MacOS? I am seeing some entries in the database that look like this:

That is a good catch -- I notice the same on my system for a few entries in the actual u8 file (e.g. those last few in the file). I have assumed they are extremely rare characters that are not covered by the unicode font I'm using -- but I haven't tested that hypothesis. There are so few, for such uncommon glosses/headwords, that I did not investigate, preferring to leave it as-is for MVP, but if you think we should modify the u8 file or investigate further with the source data I can do that.

Tanvez:
Just pulled the updated source code and ran the load file, looks like its starting to load in all the characters now!(before I canceled the script) Everything looks good to me, the phrase and character I tested previously are connecting to the lemmas.

Tanvez (approved):

arthurian (approved):
Fantastic work @elliottyates! The comments in the PR and code as well as the unit tests give a lot of confidence in the overall implementation and understanding of what's been done.

I just have one suggestion for the MVP specifically, and two general comments:

-  **Suggestion: Selecting a better font for Chinese.** On my machine, the `Songti SC` font is being used to render the glyphs, which is apparently one of the [fonts included with Mac OS Big Sur](https://support.apple.com/en-us/HT211240). I would be curious about your thoughts on [Noto Serif Simplified Chinese](https://fonts.google.com/noto/specimen/Noto+Serif+SC) and [Noto Serif Traditional Chinese](https://fonts.google.com/noto/specimen/Noto+Serif+TC), since we already use [Noto Serif](https://fonts.google.com/noto/specimen/Noto+Serif) for latin, greek, and cyrillic scripts. I don't think it would be a heavy lift to include that if you think it's better, and would be worth incorporating in the MVP. For reference, see [_fonts.scss](https://github.com/Hedera-Lang-Learn/hedera/blob/dev/static/src/scss/_fonts.scss) and the [google-webfonts-helper](https://google-webfonts-helper.herokuapp.com/fonts).
- **Comment: Loading chinese texts quicker.** On my machine, it took 6746 seconds (almost 2 hours) to load everything. That's excruciatingly slow, but I think for an MVP, we can live with it. I would rather hold off on optimizing the load until we've worked out what we're doing with the Latin/Morpheus lemmatizer, since that may serve as a model for Chinese and Russian. Ultimately, I think we may want to move the data into a separate DB that can be loaded quickly with a database dump and queried on demand to build up the lattice over time, but I think that's out of scope for the MVP.
- **Comment: Dealing with traditional vs simplified headwords.** I think for the purpose of the MVP, this will be fine even though there's a bit of redundancy. I'm not certain how best to evolve the data model to support Chinese effectively, but I would be curious if we could retain the link between the traditional and simplified characters through the existing lattice structure by creating a join node. Assuming we currently are mapping nodes like this:
    ```
    LemmaNode(我) --> LatticeNode(我)  # [wo3] /I/me/my/
    LemmaNode(個) --> LatticeNode(個)  # [ge4] /individual/this/that/...
    LemmaNode(个) --> LatticeNode(个)  # [ge4] /individual/this/that/...
    ```
    Perhaps we could update it so we get something like this instead:
    ```
    LemmaNode(我) --> LatticeNode(我)

    LemmaNode(個) --> LatticeNode(個) \  
                                      |--> LatticeNode(個 个)
    LemmaNode(个) --> LatticeNode(个) /   
    ```
    In the case of 我, there is no distinction between the traditional and simplified character (correct me if I'm wrong), but in the case of 個 and 个 there is, and we lose the link between them. The above structure would retain a distinct lemma for each character, pointing to the corresponding node in the lattice as before, but with the join node (created at load time), we would have a way of potentially switching between them. A teacher could lemmatize a text, and then use the UI to point to the parent node for both writing systems, or switch to the other one (so you see the simplified token, but then when you click on it, get the traditional).

     I suspect it's not that simple, but wanted to suggest it as a starting point. Perhaps we need a slightly different data model to support this use case, so in that case, I think we should make this a post-MVP story to revisit after getting some initial feedback.

elliottyates:
Thanks @arthurian for the comments and ideas.

# 1. Font change

Great suggestions. I was not very familiar with the CSS `font-face` rule so I learned a lot looking at this. I experimented with Noto Serif SC and TC, and in the end it seemed that, rather than trying to distinguish which to use for a given Chinese (`lang=zho`) text, we could just use the SC variety -- it [includes "Literary Chinese" support](https://fonts.google.com/noto/specimen/Noto+Serif+SC), which seems to be sufficient to render traditional characters in my test text just fine. 

Here are screenshots:


<table>
<thead>
<tr><th>Before (no SC):</th><th>After (with SC):</th></tr>
</thead>
<tbody>
<tr><td>
Screen Shot 2021-09-28 at 10 02 57 PM
</td>
<td>
Screen Shot 2021-09-28 at 10 07 26 PM
</td></tr>
<tbody>
</table>

Additional user testing may reveal a weakenss in this MVP approach; a different, probably better, approach would be to utilize the [unicode-range property](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range) of the `font-face` declaration to use SC and TC for specific Unicode ranges, and fall back to regular Noto Serif for the others. That would require working out some Unicode range maps (this is possibly simple, but not trivial), so maybe we can do this later if need be?

(Apropos of your Mac OS X observations of the font Songti SC, where did you see/discover that? Was it in your browser? When I look in the browser, I just see the default font, so I'm not sure how the browser/OS falls back to find a system font.)

# 2. Loading Chinese texts quicker

Agreed, let's do it as a follow-up to the architectural discussion; I'll hold off from creating an actual new project issue for now and see how that discussion develops, if that makes sense to you too.

# 3. Traditional vs Simplified

I'm not sure how link nodes work, so I might need an "in-person" discussion to wrap my head around this. I don't find the data model very comprehensible in general, so went with a simple approach, but if there's a better way to handle glosses than duplication, that would be awesome. I opened #343 for this, linking to your detailed suggestion in this PR -- we can use that "story" as a ["placeholder for a conversation"](https://www.agilealliance.org/user-story-conversations/), if that works?

arthurian:
The new font looks great! For reference, you can find it in the Chrome dev tools if you inspect an element and look at the **Computed** tab. It's listed at the bottom. 

hedera-chinese-rendered-font-pr349
 
I agree that it would be better to use the `unicode-range` property -- I think that's what google fonts do if you just link to them, but we ran into issues when we did that (changes we did not expect). The approach of marking the language with a class seems fine.

Thanks for creating a separate story for discussion on the traditional/simplified issue -- I think we have a path forward on a separate, but maybe related issue with #350. Definitely worth a follow-up discussion on that and the lattice.

All in all, I think this is good to merge as a first pass at implementing Chinese 👍 , with some planning to do with respect to improvements on data loading and traditional/simplified.

elliottyates:
@arthurian thanks, I had never noticed the "rendered font" there before, way down at the bottom. Good to know.

PR #361, opened 2021-09-24
Fix bug personal vocab quick add  - refactor post endpoint, test, nav bar in templates

Tanvez:
Refactor the following in relation to the quick add personal vocab feature

[ATGU-3233](https://jira.huit.harvard.edu/browse/ATGU-3233) and [ATGU-3257](https://jira.huit.harvard.edu/browse/ATGU-3257)

#355 #362

- Changed post endpoint to accommodate post without personal vocabulary list already present on the database(added lang field to post body)
- Change Vue Quick Add component to hit supported lang endpoint to populate the language options
- Added extra `lang` field to api.js and related files to account for creation of personal vocabulary list if none exist
- Added supported languages in Vuex store(state, constant, actions, api)
- Refactor some test for vue components in relation to the updated component
- Added some more test for django in relation to the changes
- Added back personal vocab in nav bar in templates. It will show up regardless if you have a personal vocab list or not, but will redirect accordingly to either the create view or list view
- Change suggestions options in quick add to show only for latin now(can change to more languages)

Notes: 
- ~supported language endpoint isnt fully integrated into vue yet just the quick add components~ Updated component(only the quick add feature) that uses static languages in constants.js to use new endpoint
- ~Most of the django templates are using the settings.LANGUAGES( except for lemmatized_text/create.html)~ Refactored lemmatized_text/create.html to use `settings.SUPPORTED_LANGUAGES`
- The vue test pass but there some errors thats logging due to child components dont have access to actions in test


dodget:
I have no comments on the specific code, but I'm not sure that we should only half integrate the new supported languages endpoint. I would consider having both `FETCH_SUPPORTED_LANG_LIST` (and the corresponding and the statically defined `LANGUAGES` in `constants.js` as tech-debt. If that could get cleaned up, that would be 💯

Tanvez:
> I have no comments on the specific code, but I'm not sure that we should only half integrate the new supported languages endpoint. I would consider having both `FETCH_SUPPORTED_LANG_LIST` and the statically defined `LANGUAGES` in `constants.js` as tech-debt. If that could get cleaned up, that would be 💯
> 
> edit: my suggestion is to supplant all uses of `LANGUAGES` from `constants.js`
Should we go ahead and replace the use of the language endpoint for now(just use constants.js) and consider that as a seperate story? or factor in the changes for the languages endpoint into this PR as well?

arthurian (approved):
LGTM :+1:

lemmatized_text/views.py
```diff
@@ -47,7 +47,13 @@ def create(request):
         lemmatized_text.clone(cloned_by=request.user)
         return redirect("lemmatized_texts_list")
 
-    return render(request, "lemmatized_text/create.html", {"select_lang": select_lang})
+    lang_dict = []
+    for lang in settings.SUPPORTED_LANGUAGES:
+        lang_dict.append({
+            "id": lang[0], "name": lang[1]
+        })
```
dodget, line 54: Rather than mutating the SUPPORTED_LANGUAGES, you could pass it directly to the template:
```py
return render(request, "lemmatized_text/create.html", {"select_lang": select_lang, "supported_lang": settings.SUPPORTED_LANGUAGES})
```

And then in the template:
```
{% for lang in supported_lang %}
    <option value="{{ lang.0 }}" {% if select_lang == lang.0 %}selected{% endif %} >{{ lang.1 }}</option>
{% endfor %}
```
You can access the list values by `.0` and `.1` in the template.

dodget (approved):
Looks good to me! Just one comment!

lemmatized_text/views.py
Tanvez: oh cool! Will make the changes now.

PR #365, opened 2021-10-01
Supported language refactor

dodget:
This PR resolves [ATGU-3276](https://jira.huit.harvard.edu/browse/ATGU-3276) by:

- Adding the `hedera.supported_languages.py` module
- Defining a `SupportedLanguage` as a namedtuple so we can use dot notation when accessing languages, as well as clearly defining the things it requires to add a supported language
- Adding a `SUPPORTED_LANGUAGES` dictionary in the module mentioned above in order to replace the definition in the settings module, as well as clearly stitching together what it takes to add a language.
- Creating language specific modules in `lemmatization.services` that house both the service and the tokenizer for the language. This is based on Elliott's work on integrating Chinese.
- Removing `lemmatization.tokenizer.py` as the parts of this module have been distributed to the concerned language specific modules

## Notes

- This work is born out of the question "If all of the current developers on this project were gone tomorrow, how can we make it simpler for a new developer to add a new supported language?"
- This was a bigger reorganization of the codebase than was initially intended. After viewing how @elliottyates clustered the Chinese service and tokenizer into a single `chinese.py` module under `lemmatization.services` I thought that would make sense for the rest of our languages as well. It would also be the most intuitive way to add a new language.
- This work represents a reorganization of code, and does not contain changes to functionality


hedera/supported_languages.py
```diff
@@ -0,0 +1,19 @@
...
+
+from lemmatization.services.chinese import ChineseService, LACChineseTokenizer
+from lemmatization.services.greek import GreekService, GreekTokenizer
+from lemmatization.services.latin import EncliticTokenizer, LatinService
+from lemmatization.services.russian import ClancyService, RUSTokenizer
+
+
+# using ISO 639.2 CODES
+
+SupportedLanguage = namedtuple("SupportedLanguage", ["code", "verbose_name", "service", "tokenizer"])
```
Tanvez, line 11: Awesome! nice to see this implemented

Tanvez (approved):
Looks good to me! Awesome refactor @dodget

arthurian (approved):
Nice work @dodget! I really like the new structure :+1:

elliottyates (approved):
Thanks, many things seem to be organized more intuitively now -- having the `Service` and `Tokenizer` in `base.py`, for instance, or the way `Lemmatizer`s are configured at `__init__()`...  and not having to import `settings` everywhere is nice too. 👍

PR #367, opened 2021-10-12
Vocab list refactor

dodget:
This PR resolves [ATGU-3255](https://jira.huit.harvard.edu/browse/ATGU-3255) by:

- Adding the `AbstractVocabList` and `AbstractVocabListEntry` abstract classes to refactor repeated fields and methods
- Adding tests to gauge the safety of the refactoring/absracting 
- Adding docstrings to functions/methods that could use some documentation

## Notes

- The refactoring/abstraction definitely helps this module to adhere to DRY principals, but I wonder if it makes it more or less comprehensible? That's an open question for me

### DIFF
--- a/vocab_list/models.py
+++ b/vocab_list/models.py
@@ -42,6 +42,10 @@ def parse_line(idx, line):
     return clean(*columns)
 
 
+def filter_lines(lines, already_exist):
+    return list(filter(lambda x: x[0] != "" and x[0] not in already_exist, lines))
+
+
 class AbstractVocabList(models.Model):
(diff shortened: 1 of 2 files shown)


arthurian (approved):
Nice work on the refactoring @dodget! I think abstracting the vocab lists is helpful overall. The only area that might be worth a little consideration before merging is around loading tab-separated data, since that's a bit harder to follow after the refactoring (mostly around the handling of the entry-specific data like the `familiarity` property).

arthurian:
The latest refactor looks excellent @dodget - the `load_tabl_delimited()` implementation is MUCH clearer. :+1:

PR #368, opened 2021-11-08
Refactor to use requirements folder instead of Pipenv

Tanvez:
- Refactor Dockerfile to use `requirements/local.txt`
- Updated README.md to reflect changes from Pipenv to requirements
- Refactor Github actions to use `requirements/local.txt`
Note: I used the following commands to export the packages into base.txt and dev.txt, this created a txt file that I think includes any dependencies required by any of the main packages:

```
pipenv lock -r --dev-only > dev.txt
pipenv lock -r > base.txt
```
I don't think anything needs to be changed for deployment since the codebuild is pulling a docker image


Tanvez:
TODO: 
- remove local and use base.txt
- test pip freeze and pip install when installing new packages
- test with removed root user

dodget (approved):
Looks good based on our convo, with the removal of the `local.txt` file

arthurian (approved):
Thanks @Tanvez. Looks good! 👍

PR #377, opened 2021-12-15
Use multistage build to help with local dev

elliottyates:
This is a PR to spark some discussion. 

For easier debugging in local development, one approach is to ~use a dev Docker image that includes dev requirements~ make use of a multistage build, where dev requirements are extra layers on top of the production image.

Pros:
- prod image can also be run locally (e.g. for `npm` service)
- dev image "tests" prod image by inheriting from it

Cons:
- when running prod image, need to target the `prod` build stage explicitly to avoid including `dev` layers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
-FROM python:3.7
+FROM python:3.7 AS prod
 
-# Upgrade pip 
+# Upgrade pip
 RUN pip install --upgrade pip
 
 # Create app working directory
@@ -27,8 +27,15 @@ RUN npm i && \
 # Copy app files
 COPY . /app
 
-# Build webpack assets 
+# Build webpack assets
 RUN npm run build
 
 # Use port 8000 for django server or 8080 for webpack dev server
 EXPOSE 8000 8080
+
+FROM prod AS dev
+
+# Install python dev dependencies
+COPY ./hedera/requirements/dev.txt /app/
+RUN pip install -r dev.txt && \
+    rm dev.txt
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,78 +39,20 @@ services:
             - "8080:8080"
         command: /bin/sh -c 'npm i && npm start'
     django:
-        image: hedera-django-dev
-        build:
-            context: .
-            dockerfile: Dockerfile
-            target: dev
+        extends:
+            file: docker-compose-common.yml
+            service: django
         container_name: hedera-django
-        working_dir: /app
-        volumes:
-            - ./cms/:/app/cms
-            - ./fixtures/:/app/fixtures
-            - ./groups/:/app/groups
-            - ./hedera/:/app/hedera
-            - ./import-data/:/app/import-data
-            - ./lattices/:/app/lattices
-            - ./lemmatization/:/app/lemmatization
-            - ./lemmatized_text/:/app/lemmatized_text
-            - ./lti/:/app/lti
-            - ./pdfservice/:/app/pdfservice
-            - ./static/src/:/app/static/src
-            - ./vocab_list/:/app/vocab_list
-            - ./manage.py/:/app/manage.py
         ports:
             - "5678:5678"  # default for debugpy
             - "8000:8000"
-        depends_on:
-            - postgres
-            - redis
-        environment:
-            RQ_ASYNC: 1
-            DATABASE_URL: postgres://hedera:hedera@postgres/hedera
-            REDIS_URL: redis://redis:6379
-            PYTHONFAULTHANDLER: 1
-            PYTHONUNBUFFERED: 1
-            PYTHONDONTWRITEBYTECODE: 1
-        networks:
-            - localdev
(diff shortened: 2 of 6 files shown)


docker-compose.yml
```diff
@@ -26,7 +26,7 @@ services:
         networks:
             - localdev
     npm:
-        image: postgres-django
+        image: hedera-django
```
elliottyates, line 29: When working on multiple projects, it's nice for the local images to include the project name.

arthurian (commented):
@elliottyates I like the idea of creating a dev image for local development that includes all of the required dev dependencies. There is a little bit of extra complexity maintaining two separate images, and we lose the simplicity of doing a simple `docker compose up`, but I think it's probably worth it. The documentation you added makes it clear enough what needs to be done. So I'm onboard with this change. :+1:

That said, one idea I'd be interested to kick around is trying to use [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) to maintain a single `Dockerfile` with named stages for prod and dev that we could then [target in docker-compose](https://docs.docker.com/compose/compose-file/compose-file-v3/#target). I haven't tried this approach, but it seems like it would give us the best of both worlds: a streamlined prod image, a dev image layered on top of prod, and the ability to simply run `docker-compose build` and `docker-compose up`.  Have you experimented with this approach in any other projects?

docker-compose.yml
arthurian: Great point.

elliottyates:
@arthurian that's a great point -- DART uses multistage builds, and used the `--target` param for `docker build` and also the `target` in docker compose. I can try to rework it. To that end, I'm not familiar with the production build and deploy process, so we'd want to make sure it was configured to only build to the prod target, not the dev target -- can you advise on that?

arthurian:
@elliottyates Awesome! In terms of the build and deploy process, you're absolutely right that we would need to make sure we're targeting the correct stage (e.g. `prod`) when doing our `docker build`. The deployment process triggers a CodePipeline that then runs CodeBuild to build and deploy the docker image, so I think we'd just need to update the `buildspec.yml` in our internal hedera infra repo. If this works well with the local dev, that should be an easy enough change to make.

docker-compose.yml
```diff
@@ -26,28 +26,40 @@ services:
...
+            - ./hedera/:/app/hedera
+            - ./import-data/:/app/import-data
+            - ./lattices/:/app/lattices
+            - ./lemmatization/:/app/lemmatization
+            - ./lemmatized_text/:/app/lemmatized_text
+            - ./lti/:/app/lti
+            - ./pdfservice/:/app/pdfservice
+            - ./static/src/:/app/static/src
+            - ./vocab_list/:/app/vocab_list
+            - ./manage.py/:/app/manage.py
```
elliottyates, line 62: I'm not a big fan of this long, needs-to-be-maintained list of mappings. It's important because mapping `node_modules` isn't a good idea between OSX and Ubuntu -- I've observed the container django blow up if it's accessing `node_modules` installed for OSX. I'm no expert here, so suggestions welcome re: `node_modules` issues.

Further, this doesn't include all the `load_*.py` scripts that are referenced in the README. If we settle on this approach, we can maybe move those into a `scripts/` dir to make it easier to include here; maybe we'll come up with a better approach to host-container mapping for development, though, and avoid that issue somehow! :)

docker-compose.yml
```diff
@@ -26,28 +26,40 @@ services:
...
     npm:
-        image: postgres-django
+        image: hedera-django
         build:
             context: .
             dockerfile: Dockerfile
+            target: prod
         container_name: hedera-npm
-        volumes:
-            - ./:/app
```
elliottyates, line 35: This container doesn't appear to need anything from the host once it's built. I suppose since it runs `npm install` on each `up` we could include `package.json` and/or `package-lock.json`. Not sure what people prefer here, but I don't think it should include `node_modules`, nor does it need access to python source code.

elliottyates:
@arthurian ready for further consideration.

Dockerfile
```diff
@@ -27,8 +27,15 @@ RUN npm i && \
...
 COPY . /app
 
-# Build webpack assets 
+# Build webpack assets
 RUN npm run build
 
 # Use port 8000 for django server or 8080 for webpack dev server
 EXPOSE 8000 8080
+
+FROM prod AS dev
```
Tanvez, line 36: Cool! Didn't we could do that with docker.

docker-compose.yml
Tanvez: We are planning to consolidate those `load_*.py` scripts in another story so that might become easier to manage.

Tanvez: Wonder if we could use something like [extends configuration](https://docs.docker.com/compose/extends/#understand-the-extends-configuration) to extend the volumes into a seperate yml file just to keep them DRY.

Tanvez (approved):
Awesome work @elliottyates !

arthurian (approved):
@elliottyates The multi-stage build looks good! 👍

docker-compose.yml
elliottyates: @arthurian interesting -- I think in this case I favor explicit, because getting tricky with volumes could lead to subtle issues which could be harder for us, let alone casual contributors, to troubleshoot. I will explore the `extends` mod @Tanvez proposed to see if it'll make documenting/maintaining clearer, though!

PR #378, opened 2021-12-15
Run python unit tests in CI

elliottyates:
Provides instructions for running unit tests locally, and adds them to CI in a separate job (it will run in parallel with the build and code quality test job).

To get tests to work, we had to ensure that local and containerized test runs have access to postgres and redis, and that the tags in our templates provided by django-webpack-loader will not result in errors at test time if webpack hasn't actually been run to build static assets. This means that the webpages rendered by Django in unit tests do not have the real JS resources in them, so this approach isn't appropriate for end-to-end testing / UI interactions, but is quicker and more focused than also having to build and serve the webpack assets for each test run.

Steps to run tests are outlined in the README.md, but to summarize:

```
docker compose up -d postgres redis
# make sure postgres and redis are available, then run:
docker compose run --rm --no-deps -e DJANGO_SETTINGS_MODULE=hedera.test_settings django python manage.py test
# feel free to bring down postgres and redis after, e.g. `docker compose stop postgres redis`
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,6 +48,34 @@ jobs:
     name: Running unit tests
     runs-on: ubuntu-latest
 
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_DB: hedera
+          POSTGRES_USER: hedera
+          POSTGRES_PASSWORD: hedera
+        # Set health checks to wait until postgres has started
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          # Maps tcp port 5432 on service container to the host
+          - 5432:5432
+      redis:
+        image: redis
+        # Set health checks to wait until redis has started
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          # Maps port 6379 on service container to the host
+          - 6379:6379
+
     steps:
         - uses: actions/checkout@v2
 
@@ -73,4 +101,4 @@ jobs:
 
         - name: Run tests
           run: |
-            python manage.py test
+            DATABASE_URL=postgres://hedera:hedera@localhost/hedera python manage.py test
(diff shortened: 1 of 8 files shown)


elliottyates:
Hmm -- I think, at this point, that I will take a break. There are a bunch of webpack errors -- not sure what the unit tests are doing here, but it seems like a complicated setup for unit tests.

hedera/tests/test_api.py
```diff
@@ -18,7 +18,8 @@
 class PersonalVocabularyQuickAddAPITest(APITestCase):
 
     def setUp(self):
-        self.created_user = User.objects.create_user(username=f"test_user{randrange(100)}", email=f"test_user{randrange(100)}@test.com", password="password")
```
elliottyates, line 21: I experienced test failures using `randrange(100)`, as sometimes generated `usernames` would actually clash, so switching to UUIDs makes things more random.

elliottyates:
This is working now, and I've updated the title and PR description, and provided explanatory comments. It's ready for review.

dodget (approved):
👍 looks good.  A nice improvement

.github/workflows/ci.yaml
```diff
@@ -42,4 +42,64 @@ jobs:
...
+
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_DB: hedera
+          POSTGRES_USER: hedera
+          POSTGRES_PASSWORD: hedera
+        # Set health checks to wait until postgres has started
+        options: >-
```
Tanvez, line 59: Cool! Didn't know there was an option for health checks.

Tanvez (approved):
Looks good to me!

arthurian (approved):
Looks good to me! Nice work dealing with the postgres/redis setup and the django-webpack-loader issues. Nice to have the python unit tests running in CI 👍

.github/workflows/ci.yaml
elliottyates: Nor I! This is cribbed from https://docs.github.com/en/actions/using-containerized-services/about-service-containers -- they have examples for postgres and redis. :)

PR #379, opened 2021-12-17
Run JS tests in CI

elliottyates:
Just adds JS tests to CI runs.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,7 +2,7 @@ name: Code Checks
 on: [push]
 jobs:
   qa:
-    name: Checking Code
+    name: Check builds and code quality
     runs-on: ubuntu-latest
 
     steps:
@@ -17,7 +17,7 @@ jobs:
           with:
             node-version: '14'
 
-        - name: Setup Cache
+        - name: Cache Python dependencies
           id: cache
           uses: actions/cache@v2
           with:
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -147,8 +147,10 @@
     },
     "collectCoverage": true,
     "collectCoverageFrom": [
-      "**/*.{js,vue}",
-      "!**/node_modules/**"
+      "**/*.{js,vue}"
+    ],
+    "coveragePathIgnorePatterns": [
+      "/node_modules/"
```
elliottyates, line 153: In practice I found that the `collectCoverageFrom` directive intended to prevent looking at `node_modules` didn't work; I found this worked better. I may have missed something, but this speeds up the coverage collection quite a bit in my tests.

.github/workflows/ci.yaml
```diff
@@ -103,3 +111,28 @@ jobs:
...
+        - name: Cache node modules
+          uses: actions/cache@v2
+          with:
+            path: ~/.npm
+            key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+            restore-keys: |
+              ${{ runner.os }}-node-
+
+        - name: Install Static Dependences
+          run: npm i --include=dev
```
arthurian, line 135: Is it my imagination or is the `--include` flag an [undocumented](https://docs.npmjs.com/cli/v8/commands/npm-install) feature of the CLI? I was just curious as I couldn't find that listed anywhere.

arthurian (approved):
Thanks @elliottyates! 👍

.github/workflows/ci.yaml
Tanvez: @arthurian I found a snippet in the `omit` section of the docs https://docs.npmjs.com/cli/v8/commands/npm-install#omit

Tanvez (approved):
Looks good to me!

.github/workflows/ci.yaml
elliottyates: Good find @Tanvez -- I used the [penultimate step of the linting job](https://github.com/Hedera-Lang-Learn/hedera/pull/379/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR49-R50) as a template for this, @arthurian, so didn't really consider the flag, as it was working; I'm guessing it is explicit about/ensures dev dependencies, even if `NODE_ENV=production`.

PR #381, opened 2022-01-06
Fix tooltip position for text familiarity dots

Tanvez:
- Removed unnecessary css in `PersonalVocab.vue`
- Updated `help-text` css in Learner.vue


arthurian (approved):
LGTM 👍

PR #382, opened 2022-01-07
Fix multiple obects returned error on lemmatized_text view

dodget:
This PR resolves [ATGU-3377](https://jira.huit.harvard.edu/browse/ATGU-3377) by:

- Making sure sub queryset contains only unique lemmatized text objects

## Notes

The filter queryset was sometimes returning the same text many times, so when `get_object_or_404` was called on that queryset with pk as the identifier, it was producing a "too many objects returned" error, as it expects only  a single text to be returned.


arthurian (approved):
LGTM 👍

Tanvez (approved):
LGTM!

PR #383, opened 2022-01-10
Add props to make quick add vocab consistant with Personal vocab sele…

Tanvez:
- Fix quick add default language to currently selected Personal Vocab tab
Note: The quick add button in the Navigation wont have access to the language from the Personal Vocab tab because the Nav is rendered using Django while the Personal Vocab is using Vue


arthurian (approved):
Looks good @Tanvez! 👍

PR #384, opened 2022-01-13
Fix webpack hot reloading

Tanvez:
- added `webpack-stats` and `/static/src` to npm volume to fix webpack hot reloading
 - added `webpack-stats` to django volume via `docker-compose-common.yml`


arthurian (approved):
Looks good! 👍

PR #385, opened 2022-01-13
Fix quick add suggested gloss so other languages can lookup headwords

Tanvez:
- removed blocking code from quick add component so other languages can lookup headwords
- Added extra parameter when looking up headwords
Note: as of right now there is no way to distinguish/match languages between glosses that is returned. Chinese is the only language that can be matched via context on the lattice node. However Russian, and Latin context are referenced with `Clancy` and `Morpheus` respectively. We would need a way to validate that the actual gloss thats returned is for the correct language.(I am not sure about Greek since there doesn't seem to be a script to load that lattice in)

### DIFF
--- a/docker-compose-common.yml
+++ b/docker-compose-common.yml
@@ -25,6 +25,7 @@ services:
             - ./lti/:/app/lti
             - ./pdfservice/:/app/pdfservice
             - ./static/src/:/app/static/src
+            - ./webpack-stats/:/app/webpack-stats
             - ./vocab_list/:/app/vocab_list
             - ./manage.py/:/app/manage.py
         depends_on:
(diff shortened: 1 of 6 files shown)


docker-compose-common.yml
```diff
@@ -25,6 +25,7 @@ services:
             - ./lti/:/app/lti
             - ./pdfservice/:/app/pdfservice
             - ./static/src/:/app/static/src
+            - ./webpack-stats/:/app/webpack-stats
```
arthurian, line 28: Nice fix! I often forget about the webpack stats.

arthurian (approved):
Looks good! 👍 

For posterity: @Tanvez and I discussed this PR in person. This tees it up for the backend to provide support for searching the lattice by language.

docker-compose-common.yml
elliottyates: (ditto! thanks!)

Issue #350, opened 2021-09-23
RFC: Morpheus data service


arthurian:
While a specific proposal is being made in this RFC, it is subject to change based on the discussion and feedback. The goal is to have consensus on the right approach.

## Summary

This is a proposal to improve Hedera's lemmatization of latin texts by implementing our own headword/lemma retrieval service rather than calling out to Morpheus via the [Perseids Morphology Service API](http://services.perseids.org/). This approach is feasible because we have a copy of the underlying data for Morpheus.

## Motivation

Two common spelling variations occur in Latin having to do with i/j and u/v. It is common to distinguish between u/v, but not i/j. Morpheus will often recognize the same word spelled differently (though not always, as shown below), but even when it's recognized, it's difficult to predict the spelling that will be used.

Examples:
- [iacio](http://services.perseids.org/bsp/morphologyservice/analysis/word?engine=morpheuslat&lang=lat&word=iacio), [jacio](http://services.perseids.org/bsp/morphologyservice/analysis/word?engine=morpheuslat&lang=lat&word=jacio) → `jacio`
- [iuvo](http://services.perseids.org/bsp/morphologyservice/analysis/word?engine=morpheuslat&lang=lat&word=iuvo), [juvo](http://services.perseids.org/bsp/morphologyservice/analysis/word?engine=morpheuslat&lang=lat&word=juvo) → `juvo`
- [iam](http://services.perseids.org/bsp/morphologyservice/analysis/word?engine=morpheuslat&lang=lat&word=iam), [jam](http://services.perseids.org/bsp/morphologyservice/analysis/word?engine=morpheuslat&lang=lat&word=jam) → `iam`
- [adiudicare](http://services.perseids.org/bsp/morphologyservice/analysis/word?engine=morpheuslat&lang=lat&word=adiudicare), [adjudicare](http://services.perseids.org/bsp/morphologyservice/analysis/word?engine=morpheuslat&lang=lat&word=adjudicare) → `adjudico`, [adivdicare](http://services.perseids.org/bsp/morphologyservice/analysis/word?engine=morpheuslat&lang=lat&word=jacio) → _None_
- [nuntius](http://services.perseids.org/bsp/morphologyservice/analysis/word?engine=morpheuslat&lang=lat&word=nuntius) → `nuntius` [nuncius](http://services.perseids.org/bsp/morphologyservice/analysis/word?engine=morpheuslat&lang=lat&word=nuncius) → _None_

[...]

PR #387, opened 2022-02-03
Switch to local headword retrieval for latin

arthurian:
This PR updates the `LatinService` to use a local headword retrieval strategy rather than calling out to the Perseids Morphology Service (e.g. `MorpheusService`). This is a first step along the path to improving lemmatization and dealing with spelling issues since it gives us more control over the entire process (see also: #350).

Notes:
- Added `latin_lexicon.csv` which maps tokens to lemmas. This is an export of data provided to us by Logeion (`LatinMorph16.db`).
- Added `LatinLexicon` model to `lemmatization.models`. Note that this model includes just the `token` and `lemma` fields from the `LatinMorph16.db`. A `frequency` field has also been added in anticipation of adding frequency data to improve the order of the results when there are multiple matching lemmas, although we don't have that frequency data yet.
- Added `load_latin_lexicon.py` to load the data into `LatinLexicon`, with some minor optimization to do bulk inserts. This is an isolated table that has no relationships with any others.
- Slightly refactored the `services.base` module to include the original `MorpheusService` class, which is now only used by the `GreekService`. 
- Updated `LatinService` to query `LatinLexicon` directly. Note that it matches on the `token` field exactly (for now).
- Added basic unit test in `lemmatization.tests` to exercise the `LatinService`.

@bilbe @Tanvez


Tanvez (approved):
Looks good @arthurian! tests and import works great!

lemmatization/services/latin.py
```diff
@@ -34,8 +36,30 @@ def re_tokenize_clitics(tokens):
...
+    def lemmatize(self, form: str) -> List[str]:
+        """
+        Retrieves matching lemmas (headwords) for a word by querying our local database of
+        latin tokens and lemmas. This was previously a remote call to the remote
+        Perseids Morphology Service (e.g. MorpheusService).
+
+        Results are returned in order of highest frequency first, or simply
+        alphabetical by lemma if there is no frequency data.
+        """
+        results_qs = LatinLexicon.objects.filter(token=form)
```
arthurian, line 60: Should this be a case sensitive search or case insensitive?

IvyJL:
Could we try case sensitive?
Right now I think we are ignoring case somewhere in the process, but in would be nice to distinguish if there are proper names to be found. 
For example, quintus means “fifth” but Quintus is also a male first name. 
>

arthurian:
@IvyJL Yes, absolutely! @bilbe and I were just talking about this and he mentioned something similar. We'll try it case-sensitive for now.

bilbe:
The problem with case sensitive is what about capitalization at the beginning of a sentence or line? If Quintus is capitalized as the first word of a sentence, which is it? Maybe it would be better to give people the option to disambiguate. My two cents.

IvyJL:
Oh, yes, sure. Users may have to disambiguate. But if we could retrieve both upper and lower case so that they could disambiguate, that would be good. Instead of just ignoring all uppercase. I plan to give as best practice in the instructions to only capitalize proper names, which some textbooks do anyway.

PR #396, opened 2022-03-10
add set to filter out dupes in load_tab_delimited

Tanvez:
- Used `set()` to filter out duplicates in the lines(parsed tab delimited file) to filter out duplicates that was causing the vocabulary upload to fail.
- Refactor the `load_tab_delimited` function to filter `headword` and `gloss`
- Commented out the `unique_together` flag in the `PersonalVocabularyListEntry` and `VocabularyListEntry`
- Added migration files for model changes
- Test for duplicate entries in personal and vocab entries


bilbe:
Looks great, Vesna! 👍

vocab_list/tests.py
```diff
@@ -17,8 +17,17 @@
 bin\tbash
 hello\tworld"""
 
+dupe_vocab_list_tsv_content = b"""\
+car\tarmor
+car\tarmor"""
+
+dupe_headword_list_tsv_content = b"""\
+nintendo\tmario
+nintendo\tmario2"""
```
arthurian, line 26: It might be instructive to use `quam` for this test since it's a real example of a word with different definitions:

- **quam** as possible as
- **quam** how
- **quam** than

vocab_list/models.py
```diff
@@ -198,7 +202,12 @@ class Meta:
         verbose_name = "vocabulary list entry"
         verbose_name_plural = "vocabulary list entries"
         order_with_respect_to = "vocabulary_list"
-        unique_together = ("vocabulary_list", "headword")
+        # commented out due to failing uploads with same headword but different definitions
+        # Example:
+        # quam - as possible as
+        # quam - how
+        # quam - than
+        # unique_together = ("vocabulary_list", "headword")
```
arthurian, line 210: I'd suggest deleting the `unique_together` line rather than just commenting it out. Its disappearance is preserved in the history as well as the migration file.

The comment is great, though, and might be worth using as the basis for documenting our assumptions about how the model will be used. Since the `VocabularyListEntry` and `PersonalVocabularyListEntry` are both derived from the same base class `AbstractVocabListEntry`, it might be worth moving the comment there. You could even include it as part of a docstring for the model class. Maybe something like this?

```python
class AbstractVocabListEntry(models.Model):
    """
    An abstract vocabulary list entry class.

    Entries should have a **headword** and **gloss** (definition), and are expected 
    to be linked into the lattice. 

    Note that a headword may be repeated in a list if it has different definitions. For example,
    the latin word quam may be included three times:

        quam - as possible as
        quam - how
        quam - than
    """

```

vocab_list/models.py
```diff
@@ -222,7 +231,8 @@ class Meta:
         verbose_name = "personal vocabulary list entry"
         verbose_name_plural = "personal vocabulary list entries"
         order_with_respect_to = "vocabulary_list"
-        unique_together = ("vocabulary_list", "headword")
+        # see comment for VocabularyListEntry
+        # unique_together = ("vocabulary_list", "headword")
```
arthurian, line 235: Same comment here about deleting `unique_together`. A comment probably isn't needed here if the assumptions are documented in the base class.

vocab_list/models.py
```diff
@@ -73,8 +73,12 @@ def load_tab_delimited(self, fd, entry_model, extra_attrs={}):
         Returns: bulk create query of vocab list entries
         """
         lines = [parse_line(idx, line) for idx, line in enumerate(fd)]
-        existing_headwords = entry_model.objects.filter(vocabulary_list=self).values_list("headword", flat=True)
-        new_headwords = filter(lambda x: x[0] != "" and x[0] not in existing_headwords, lines)
+        # filter out duplicates in the files
+        # how do we know its not a real duplicate vs word duplicate with different definitions? sets!
+        filter_repeats_lines = list(set(lines))
```
arthurian, line 78: This is a nice approach to eliminating duplicates. Just a note that `set()`works because `lines` consists of a series of 2-tuples returned by `parse_line()` and tuples are immutable/hashable. If that implementation changes in the future (e.g. the tuple is changed to a mutable list), then this could break. 

Compare:

```python
>>> set([("quam", "as possible as"), ("quam", "how"), ("quam", "how")])
{('quam', 'how'), ('quam', 'as possible as')}
>>> set([["quam", "as possible as"], ["quam", "how"], ["quam", "how"]])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'list'
```

The good thing is we have some tests, so that should be caught if it happens.

arthurian (approved):
Looks good @Tanvez! A few suggestions, but otherwise good to go. Nice work on the unit tests. 👍

vocab_list/models.py
Tanvez: Ohh i see! Good to know for the future! So data structure that is considered mutable(list, dict) would not work in a set. Ex: list of dicts
```python
set([{"id": 1}, {"id": 1}, {"id": 2}])
```

vocab_list/models.py
Tanvez: Makes sense! will update with the docstring.

PR #398, opened 2022-03-24
Lattice cleanup

arthurian:
This PR cleans up some unused lattice models and functions to prepare for some updates to the lattice/lemmatization process.

Notes:
- Changed `lookup_form(s)` to `looup_form(s, lang)` because we only want candidate lemmas for the target language. Also removed unused options to filter by `context` or `annotate` the results.
- Remove `lattices.models.FormNode` as it is not being used. Also removed reference from LatticeNode via `form_strings`. 
- Moved `LatinLexicon` model into the `lemmatization.services.latin` since it is specific to that language service. This is how we expect other languages to do it as well (e.g. if `russian.

Todos:
- [x] Check that migrations work as expected after moving `LatinLexicon`.
- [x] Fix broken lemmatization tests.

@bilbe @Tanvez


Tanvez (approved):
LGTM!

PR #400, opened 2022-04-19
Add frequency data

Tanvez:
- Refactor load_latin_lexicon script


arthurian:
@Tanvez should this PR be closed since we're not using the `LatinLexicon` model anymore? Or does it need to be updated in some way to fit the new model?

PR #405, opened 2022-06-09
This solves the problem with all words being underlined in red (inste…

bilbe:
…ad of just the words with the same lemma) as well as all words being read because the they weren't being seen as having no lemmas.

--Bill B.


arthurian:
Resolved conflicts and going to go ahead and merge these UI fixes.

PR #408, opened 2022-08-08
Load latin form to lemma

Tanvez:
This PR adds load scripts for `formtolemma`, `lemma` tables and a merge script to outer join `distinct_lemmas.csv`, `shortdefs.csv`, `frequencies.csv`, and `ivy_lattice_new.csv`(converted from TSV file) Story: [ATGU-3628](https://jira.huit.harvard.edu/browse/ATGU-3628) and [ATGU-3643](https://jira.huit.harvard.edu/browse/ATGU-3643)

- Added load script `load_latin_lemma.py` to load in using refactor lemma model
- Added load script `load_latin_formtolemma.py` to load in forms to reference existing lemmas
- Added files required to create the master file with `lemma`, `def`, `alt_lemma`, `label`, `rank`, `count`, `rate` properties
- Added merge script `merge_lemma_and_frequency.py` to outer join csv files
- Updated `README.md` with new load script instructions
Note: gloss is not populated from these scripts, a separate script will be needed to create those entries


arthurian (approved):
Overall looks good @Tanvez! Just a few minor suggestions.

load_latin_lemma.py
```diff
@@ -0,0 +1,56 @@
...
+        else:
+            data["count"] = 0
+        if rate:
+            data["rate"] = float(rate)
+        else:
+            data["rate"] = 0.000
+        if not label:
+            data["label"] = lemma.lower()
+        # only adds to the batch lemma does not already exist in the database
+        if not lemma_exist and lemma:
```
jaguillette, line 48: Probably not a big deal, but maybe this check could happen before the other fields get checked and set above? That way if it's not going to get added to the batch it doesn't get processed. Might shave some time off at scale.

merge_lemma_and_frequency.py
```diff
@@ -0,0 +1,30 @@
...
+merge_lemmas_short_defs_frequencies = pd.merge(merge_distinct_lemmas_short_defs, lemma_frequencies, on="lemma", how="outer")
+merge_lemmas_short_defs_frequencies.set_index("lemma", inplace=True)
+
+final_aggregated_lemmas = pd.merge(merge_lemmas_short_defs_frequencies, ivy_lattice, on="lemma", how="outer")
+
+# lowercase capital letters
+final_aggregated_lemmas["lemma"] = final_aggregated_lemmas["lemma"].str.lower()
+
+# fill null values with blank
+final_aggregated_lemmas[["def"]] = final_aggregated_lemmas[["def"]].fillna("")
```
jaguillette, line 25: These work as is, but if you want you can access the columns with just `final_aggregated_lemmas.lemma` or `final_aggregated_lemmas.def` since the column names are good.

jaguillette (approved):
Looks good to me! I left a couple stylistic comments, but everything seems good.

load_latin_lemma.py
Tanvez: Yea, definitely makes sense to do the operations only if its being added to the batch.

load_latin_form_to_lemma.py
```diff
@@ -0,0 +1,37 @@
...
+    all_lemmas_qs = Lemma.objects.all()
+    lemmas_data = {q.lemma: q for q in all_lemmas_qs}
+    count = 0
+    batch_size = 5000
+    batch = []
+    all_formtolemma_qs = FormToLemma.objects.values()
+    formtolemma_data = {q["form"]: q for q in all_formtolemma_qs}
+    for row in tqdm(csv_reader, total=total_lines):
+        count += 1
+        val = list(row.values())
```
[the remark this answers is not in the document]
Tanvez: Oh I didn't know that will definitely update the code to guard against mismatch column order

lemmatization/models.py
```diff
@@ -71,12 +72,12 @@ def load_tab_delimited(fd, context, lang):
                 form=form,
                 lemma=lemma,
             )
-            FormToLemma.objects.get_or_create(
-                context=context,
-                lang=lang,
-                form=strip_macrons(form),
-                lemma=strip_macrons(lemma),
-            )
+            # FormToLemma.objects.get_or_create(
```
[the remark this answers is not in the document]
Tanvez: Will go ahead and remove!

PR #416, opened 2022-09-02
add gloss loading script

Tanvez:
This PR adds loading script to load in definitions into the `gloss` table
Note: there are some changes for words/normalized_words with capitalization e.g. `Sum` - would not be recognized as a lemma and would return no glosses.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,7 @@ $ docker compose exec django python manage.py reset_db --no_input
 # Run commands to reload the data
 $ docker compose run --rm django python manage.py shell -c "import load_latin_lemma"
 $ docker compose run --rm django python manage.py shell -c "import load_latin_form_to_lemma"
+$ docker compose run --rm django python manage.py shell -c "import load_latin_gloss"
 ```
 
 ## Adding a new language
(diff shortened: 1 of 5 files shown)


jaguillette (approved):
Looks good to me! I thought about just doing `.lower()` on all the strings, just in case, but I did a timeit test that convinced me you have the right idea avoiding that overhead at the scale of the number of words likely to be imported:

```python
>>> import timeit
>>> timeit.timeit('somevar = "whatever"', number=10000)
0.00026605800667311996
>>> timeit.timeit('somevar = "whatever".lower()', number=10000)
0.0011392580054234713
```

Tanvez:
> Looks good to me! I thought about just doing `.lower()` on all the strings, just in case, but I did a timeit test that convinced me you have the right idea avoiding that overhead at the scale of the number of words likely to be imported:
> 
> ```python
> >>> import timeit
> >>> timeit.timeit('somevar = "whatever"', number=10000)
> 0.00026605800667311996
> >>> timeit.timeit('somevar = "whatever".lower()', number=10000)
> 0.0011392580054234713
> ```

We could definitely add that to the import script since we wouldn't be importing this many new glosses frequently. I can see a case where the database contain a capitalized entry and a lower case entry was being looked up and vice versa(noted in the PR). I guess it would depend on how we want to sanitize the data/words. Thoughts @bilbe ?

arthurian (approved):
Looks good @Tanvez ! 👍

PR #420, opened 2022-09-08
Update gloss loading script

arthurian:
This PR updates `load_latin_gloss.py` script to improve the load time since it can take >10 mins to load a remote DB.

Notes:
- Replaced all database queries inside the loop with dict lookups. This produced a significant speed up and brought the load time to under a minute.
- The `all_gloss_data` dict lookup is intended to check if a gloss exists in the DB, which means we need the lemma ID as part of the lookup key.
- Only 16,340 (31%) of the 51,374 entries in `distinct_lemma_short_defs_frequencies.tsv` have associated glosses. Once loaded, the gloss table should have >= 16,340 glosses since a gloss may be used more than once. 
- After an initial load, the gloss table has 16,718 rows.

### DIFF
--- a/load_latin_gloss.py
+++ b/load_latin_gloss.py
@@ -15,13 +15,20 @@ with open(file_path) as f:
     batch = []
     all_lemmas_qs = Lemma.objects.all()
     all_gloss_qs = Gloss.objects.all()
+    lemmas_data = {q.lemma: q for q in all_lemmas_qs}
+    all_gloss_data = {str(g.lemma_id) + g.gloss: g for g in all_gloss_qs}
+
     for row in tqdm(csv_reader, total=total_lines):
         lemma = row["lemma"]
         gloss = row["def"]
-        lemma_exist = all_lemmas_qs.filter(lemma=lemma).first()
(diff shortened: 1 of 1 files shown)


jaguillette (approved):
LGTM, thanks!

load_latin_gloss.py
```diff
@@ -15,13 +15,20 @@
     batch = []
     all_lemmas_qs = Lemma.objects.all()
     all_gloss_qs = Gloss.objects.all()
+    lemmas_data = {q.lemma: q for q in all_lemmas_qs}
+    all_gloss_data = {str(g.lemma_id) + g.gloss: g for g in all_gloss_qs}
```
elliottyates, line 19: @arthurian not sure what effect it would have on speed, but you could offload some memory requirements onto the ORM using:

```py
    all_lemmas_qs = Lemma.objects.all().values()  # need all fields on line 31
    all_gloss_qs = Gloss.objects.all().values('lemma_id', 'gloss')  # maybe only need these two fields?
```

[Using `values()` would return `dict`s instead of entire ORM objects](https://docs.djangoproject.com/en/dev/ref/models/querysets/#django.db.models.query.QuerySet.values), which might give you a speed and memory boost as well.

PR #421, opened 2022-09-14
remove lattice from vocab models

jaguillette:
Removes the lattice elements from vocabulary entry and list models and starts to fix some of the subsequent errors in vue files and elsewhere in django. This is unlikely to be comprehensive, but gets us starting from the same place with the new model.


arthurian (approved):
Thanks @jaguillette! Note that it looks like there are a few linting warnings that might need to be fixed, but otherwise looks good to merge so we're starting with the updated vocab models.

Tanvez (approved):
Looks good to me! @jaguillette

PR #423, opened 2022-09-19
Fix vocab known words

Tanvez:
This PR fixes some of the bugs with migrating to the new lemma model.

- [x] Fix learner vue vocab stats(unique tokens wrong number) and updated the related stats model method
- [x] Fix learner view highlighting lemmatized text
- [x] Fix correct gloss to show in personal vocab view
- [x] Fix import of personal vocab to link to correct lemma
- [x] Fix adding personal vocab/familiarity via learner view
- [x] Show single system gloss in learner vue(first indexed gloss in the glosses array)
- [x] Fix personal vocab table to show lemma and formtolemma(headword)
- [x] fix javascript console error for rank in learner.vue
Notes: I applied some styling to learner.vue so it was easier to read - we might need to add a styling suite later down the road
TODO: We still need to update the vue and django tests, and fix exports with macrons


arthurian (approved):
Looks good @Tanvez (reviewed synchronously with @jaguillette). :+1:

jaguillette (approved):
LGTM!

PR #424, opened 2022-09-22
Unlattice vocab list

jaguillette:
This PR removes the lattice framework from the core functionality of vocab lists and updates references to `gloss` in vocab list entries to be called `definition`, to separate them from proper `gloss`es that exist as database entities. To that end:
- Many instances of `gloss` have been replaced with `definition` where appropriate in Vue components.
- References to lattice nodes have been replaced with references to lemmas

In addition, this PR hardens the vocab list import and export processes. The meat of these changes are in the vocab list models, where the import process is handled.

### Import

**Old Behavior:** When importing, the old behavior was to expect a tab-separated file with two columns, the first having headwords and the second having definitions. Additional columns would break the import. The import page just said that a tab separated file was expected.

**New Behavior:** The import still expects a tab separated file, but now requires headers from the database field names rather than fixed column positions. Any field on the vocab entry data model is valid for upload, including foreign key relationships when they have `_id` appended to them, so for the `lemma` foreign key relationship the column header would be `lemma_id`. Any columns that don't conform to this are ignored. This allows for familiarity scores and established lemma connections to be imported as well as headwords and definitions.

### Export

**Old Behavior:** When exporting, you get a comma-separated spreadsheet with the headword, definition, and familiarity score (if applicable). This spreadsheet included column headers for these fields.

**New Behavior:** When exporting, you get a tab-separated spreadsheet with headword, definition, familiarity (if applicable), and lemma_id. This is consistent with the import format, such that any vocabulary list that you export should be able to be imported with all information intact, including familiarity ratings and any lemma disambiguation that has been done manually (no interface for it yet, but thinking ahead).

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,3 +251,20 @@ $ docker compose run --rm django python manage.py shell -c "import load_latin_gl
 ## Adding a new language
 
 In order to add a new language, you will need to add an entry to `SUPPORTED_LANGUAGES` in `hedera.supported_languages.py`. Every language requires a "code", "verbose_name", "service", and "tokenizer." Language specific services are defined in `lemmatization.services`. Please, review the existing services and tokenizers, and add your language's module.
+
+## Debugging
+
+To debug from within vscode you can enter this line into the file/endpoint you want to debug. [Reference to debugger](https://code.visualstudio.com/docs/editor/debugging) and [debugpy vscode reference](https://code.visualstudio.com/docs/python/debugging)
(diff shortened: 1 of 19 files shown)


arthurian (approved):
This looks great @jaguillette! (We just reviewed synchronously - follow-up from yesterday) :+1:

Tanvez (approved):
Looks great! @jaguillette

PR #425, opened 2022-09-27
Disambiguation UI Updates

arthurian:
This PR updates the lemmatized text instructor view so that disambiguation actions are persisted to the database and displayed correctly on the UI. 

Changes:
* Fixed the `FormDisambiguation` Vue component so that selecting a different lemma will save that change to the database immediately. It will update the specific token with the associated lemma ID and gloss IDs.
* When selecting a different lemma, all available glosses will be selected by default.
* Within a particular lemma, glosses can be toggled on/off.
* Added a minor sanity check to the enclitics lemmatizer.


Tanvez (approved):
LGTM!

jaguillette (approved):
LGTM!

arthurian:
@Tanvez @jaguillette and I reviewed synchronously. Moving ahead with merge.

PR #426, opened 2022-09-29
Add pdf-service to docker compose and fix handout

arthurian:
This PR adds the `pdf-service` to docker, which is required to generate a handout in the learner view, and fixes the handout using the new models. Note that this piece can probably be replaced with a simpler method of generating PDFs, but for now just trying to align with what is running in dev/prod.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,7 +27,17 @@ services:
             - localdev
         ports:
             - "6379:6379"
-
+    pdfservice:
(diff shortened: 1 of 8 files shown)


Tanvez (approved):
Looks good @arthurian !

arthurian:
Thanks @Tanvez! Going to go ahead and merge since we reviewed/discussed this synchronously earlier today.

PR #430, opened 2022-10-04
fix vocab list not loading in text ui

Tanvez:
Fix zero division when calculating `link_status`


arthurian (approved):
Looks good to me @Tanvez. :+1:

PR #431, opened 2022-10-06
fix quick add

jaguillette:
This commit fixes the quick add component to work without the lattice. Quick add prompts users for a headword and definition, and as they type the headword, a set of radio button options appears with lemmas to connect to. The selected lemma is linked to the vocab list entry when it is added to the list.

This commit also makes some changes to the api and models in django. This improves error logging on the quick add endpoint and fixes an error that would occur when the `data` method was called on vocab list entries that didn't have an associated lemma.

### DIFF
--- a/hedera/api.py
+++ b/hedera/api.py
@@ -1,4 +1,8 @@
 import json
+import os
+import re
+import sys
+import traceback
 
 from django.db.models import Q
 from django.http import Http404, JsonResponse
@@ -386,7 +390,13 @@ class PersonalVocabularyQuickAddAPI(APIView):
         return lang_list
 
     def check_data(self, data):
-        keys = ["familiarity", "headword", "definition", "vocabulary_list_id"]
+        """Ensures that all expected keys are in the data."""
+        keys = [
+            "familiarity",
+            "headword",
+            "definition",
+            "vocabulary_list_id"
+        ]
         for key in keys:
             if key not in data:
                 return False
@@ -397,8 +407,8 @@ class PersonalVocabularyQuickAddAPI(APIView):
         checked_data = self.check_data(data)
         if checked_data is not True:
             return JsonResponseBadRequest({"error": "Missing required fields"})
-        if "lemma" in data and data["lemma"] is not None:
-            data["lemma"] = get_object_or_404(Lemma, pk=data["lemma"])
+        if "lemma_id" in data and data["lemma_id"] is not None:
+            data["lemma"] = get_object_or_404(Lemma, pk=data["lemma_id"])
         try:
             # to handle new quick add when list doesnt exist create using lang if
             if data["vocabulary_list_id"] is None:
@@ -413,7 +423,15 @@ class PersonalVocabularyQuickAddAPI(APIView):
             return JsonResponse({"data": {"created": True, "data": new_entry.data()}})
 
         except Exception as e:
-            return JsonResponseBadRequest(data={"error": f"{e}"})
+            exception_message = str(e)
+            exception_type, exception_object, exception_traceback = sys.exc_info()
+            filename = os.path.split(exception_traceback.tb_frame.f_code.co_filename)[1]
+            return JsonResponseBadRequest(data={
+                "error": f"{exception_message}",
+                "error_type": exception_type.__name__,
+                "error_traceback": traceback.format_tb(exception_traceback),
+                "error_filename": filename
+            })
 
 
 class JsonResponseBadRequest(JsonResponse):
--- a/hedera/api.py
+++ b/hedera/api.py
@@ -174,10 +174,13 @@ class LemmatizationFormLookupAPI(APIView):
     def get_data(self):
         form = self.kwargs.get("form")
         lang = self.kwargs.get("lang")
+
         # gets list of forms
         forms = FormToLemma.objects.filter(lang=lang, form=form)
         if not forms:
-            forms = FormToLemma.objects.filter(lang=lang, form=form.lower())
+            language_service = SUPPORTED_LANGUAGES[lang].service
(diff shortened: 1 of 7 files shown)


hedera/api.py
```diff
@@ -413,7 +422,15 @@ def post(self, request, *args, **kwargs):
...
         except Exception as e:
-            return JsonResponseBadRequest(data={"error": f"{e}"})
+            exception_message = str(e)
+            exception_type, exception_object, exception_traceback = sys.exc_info()
+            filename = os.path.split(exception_traceback.tb_frame.f_code.co_filename)[1]
+            return JsonResponseBadRequest(data={
+                "error": f"{exception_message}",
+                "error_type": exception_type.__name__,
+                "error_traceback": traceback.format_tb(exception_traceback),
+                "error_filename": filename
```
Tanvez, line 432: Nice update to error logging!

static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
```diff
@@ -231,32 +239,55 @@
...
+        // Don't do anything if headword is empty
+        if (this.headword === '') {
+          return null;
+        }
+
+        // Get headword from database if it isn't already in state
+        if (!Object.hasOwn(this.$store.state.forms, this.headword)) {
+          const newLemma = await this.$store.dispatch(FETCH_LEMMAS_BY_FORM, {
+            form: this.headword,
+            lang: this.vocabularyListItem.lang,
```
Tanvez, line 262: Is `newLemma` being using anywhere?

Tanvez (approved):
Awesome fix @jaguillette!

bilbe:
Not really working for me. Maybe we should look at it together.

bilbe:
Everything is working much better now!

However, part of my problem was that I wasn't expecting only exact matches to yield results and there is also an issue with the macrons. In the end I'd like to see typing a word to produce a visible autocomplete-like list that you can just choose from.

Here's a words that produces interesting results: _auxilium_
typing along:
_au_ only produces one result _au_ the interjection
_aux_ = nothing
_auxi_ = _augeo_ (which has a third principle part that is an exact match)
_auxil_ = nothing
_auxili_ = _auxilium_ (I guess because it's second principle part is _auxiliī_ and the macron'ed _i_ on the end is not considered part of the word)
_auxiliu_ = nothing
_auxilium_ = _auxilium_
ALSO
_auxilii_ = _auxilium_
BUT
_auxiliī_ = nothing (although is should match _auxilium_)

static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
```diff
@@ -224,39 +221,62 @@
...
         this.showSuccesAlert = false;
-        this.showUnsuccessfullAlert = false;
+        this.showUnsuccessfulAlert = false;
         this.$store.dispatch(RESET_LATTICE_NODES_BY_HEADWORD);
       },
+      /*
+      Get the lemma options from the database by looking up the provided
+      headword as a lemma form. Runs on input to the headword field.
+       */
       async getHeadword() {
```
arthurian, line 241: This might be a good candidate to `debounce()` so that way multiple calls aren't triggered until after the user is finished typing (possible future enhancement). There's an example of this here:

https://github.com/Hedera-Lang-Learn/hedera/blob/2c5dfdba41301e0921c67deca87a4afde17f99ab/static/src/js/app/modules/LemmatizedText.vue#L86

static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
```diff
@@ -295,12 +315,27 @@
         });
         return formatedNodes;
       },
+      // gives access to the state store of forms
+      stateForms() {
+        const { forms } = this.$store.state;
+        return forms;
+      },
       showLatticeNodeList() {
```
arthurian, line 323: Note that this can probably be removed now as well as the `latticeNode()` method above.

arthurian (approved):
Looks great @jaguillette! The UI seems to work well and is intuitive, and the improved error handling is nice. :+1:

static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
```diff
@@ -180,29 +176,30 @@
...
           vocabularyListId: vocabularyListItem.id,
           familiarity: this.familiarityRating,
           lang: vocabularyListItem.lang,
-        });
+          lemmaId: null,
+        };
+        if (this.lemmaId) {
+          newEntryData.lemmaId = this.lemmaId;
+        }
+        console.log(newEntryData);
```
Tanvez, line 191: Think we should remove the console logs.

Tanvez (approved):
Looks good to me! @jaguillette

static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
jaguillette: This seems like a good idea, but my naive attempt to implement it quickly resulted in getHeadword not running at all, so I'll add a story to the backlog for it. I don't think it'll take much to do, but I want to focus on getting stuff together for this afternoon's meeting.

static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
jaguillette: nixed!

static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
jaguillette: good catch, thanks!

PR #433, opened 2022-10-18
Edit personal vocab

Tanvez:
This PR adds editing options to the personal vocabulary list.

## Changes:

- `npm` packages have been updated with bootstrap-vue and its dependencies
- `toast` notifications for successful/failing personal vocab entry updates
- auto-complete for `Lemma` edits which includes the joined glosses in the suggestions dropdown
- `PersonalVocabularyListAPI POST` endpoint is pullling field names from the model and matching it with the data payload when editing an entry
- Fix hidden tooltip for familiarity in the vue-good-table component
- Fix an error for vocab list not loading
## Notes: 
- when editing the entries the order should not change once the edit has been saved to the database
- autocomplete will perform lookup based on the `startswith` functionality - this might be costly and slow down since the database does `n` look ups(not a single lookup)
*** edited - added another match filter in conjunction with `startswith` since there are forms with like `ad` that would not show up with only `startswith`

## TODO/Future improvements:
- add test
- make autocomplete modular/component based for reusability
- reduce lookup operations in database


arthurian (approved):
Overall this looks good @Tanvez. Nice work getting the autocomplete functionality working :+1:. As a concrete follow-up to merging this PR, we should add some unit tests to exercise the `PartialMatchFormLookupAPI` and `PersonalVocabularyListAPI` (as noted in your comments).

jaguillette (approved):
Looks good to me! Left a couple notes, but nothing blocking.

PR #427, opened 2022-09-30
Load shelmerdine and DCC latin vocab

arthurian:
This PR fixes and combines the shelmerdine and DCC vocab load scripts

## Changes

- Combined `shelmerdine_load.py` and `dcc_latin_morpheus_load.py` into one script that loads the system-wide (core) vocabulary lists. 
- Removed `dcc_load_old.py` since it is deprecated.
- Updated `README.md` with an instruction to load the core vocab after the lemma, form2lemma, and gloss tables are populated.

## Notes

- Vocab list entries should be linked to lemmas automatically when they are saved (this was implemented in #424), so it's not necessary to do that explicitly in the load script anymore. Also note that we are not creating lemma or form2lemma entries for any vocab item. The assumption is that the forms and lemmas must already exist in the database.
- The source data (CSV files) are not changed in any way for this PR -- we are just fixing the existing load scripts. That said, it may be worth converting CSV->TSV and having an alternate version compatible with the new upload functionality. 
- It's unclear what is different between the different versions of the DCC vocab (`dcc_latin_morpheus.csv`, `dcc_latin_morpheus.orig`, `dcc_latin_morpheus.tab.txt`) and which one we should be using. Since the old script used `dcc_latin_morpheus.csv`, that's the one we're using here as well. We should be documenting these differences if we're going to have more than one version of a vocabulary list and it might also be helpful to note the source/provenance of that list.
- While there's only a single version of the Shelmerdine vocab (`shelmerdine.csv`), it's missing column headers which would be useful to have. It appears that it's `headword`, `chapter`, `definition`, but I'm just guessing about the `chapter` column. Also, where did this list come from?
- The failing code checks on this PR are about the vocab unit tests, which need to be fixed separately.

### Shelmerdine Vocabulary

After loading the `shelmerdine.csv` vocabulary, 23 of 695 items were unable to be linked to lemmas:

```sql
SELECT ve.headword, ve.definition
FROM vocab_list_vocabularylistentry ve
JOIN vocab_list_vocabularylist vl on (ve.vocabulary_list_id = vl.id)
WHERE vl.title = 'Shelmerdine 2nd Edition Vocabulary' and ve.lemma_id is null;
```

| headword | definition |
| :--- | :--- |
| temptō, temptāre, temptāvī, temptātus | to try, attempt; test, prove |
| Trōia, Trōiae, f. | Troy |
| Trōiānus, Trōiāna, Trōiānum | Trojan |
[...]

PR #434, opened 2022-10-19
Update latin vocab

arthurian:
This PR updates the vocab import data for latin to facilitate the import of the system vocab lists with some additional cleanup included as well. Note that this is a follow-up to #427.

Changes:
- Converted the original  [DCC Core Vocabulary](https://dcc.dickinson.edu/vocab/core-vocabulary) and [Shelmerdine Vocabulary](https://hackettpublishing.com/classics/classical-languages/latin/introduction-to-latin-second-edition) CSV files to TSV since this is our target format for import. Older variants have been pruned, so now we're just keeping the original CSV and the new TSV. 
- The  `dcc_latin_morpheus.tsv` and `shelmerdine_latin.tsv` have been prepared so they can be imported directly using the UI or the `load_latin_vocab.py` script. This is a one-time action when a new instance is being setup and is noted in the README already. If the UI is used, it's necessary to change the *owner* of the vocab list to `None` so that they show up as "system" vocab lists. 
- Removed `latin_lexicon.csv` and `latin_lexicon_frequencies.csv` since these are deprecated in favor of  `distinct_lemma_short_defs_frequencies.tsv` and `form_to_lemma.tsv`.
- Renamed `wonky.tsv` to `ivy_wonky.tsv` so it's clear where the data came from.


jaguillette (approved):
Looks good to me! The new filenames being more explicit is great, and happy to have things pruned that aren't in use.

Tanvez (approved):
LGTM!

PR #435, opened 2022-10-21
Remove duplicates in formtolemma

arthurian:
This PR removes some duplicate entries that were causing foreign key errors on import.

For example:

```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "unique_lang_form_lemma"
DETAIL:  Key (lang, form, lemma_id)=(lat, visites, 50909) already exists.
```

@bilbe does this look OK to you? I'm assuming they were just erroneous entries.


bilbe:
Yes, looks good! Thanks!

bilbe (approved):
Yes, this looks great!

PR #438, opened 2022-10-21
Change to use new gloss file

Tanvez:
Added new gloss file @bilbe provided and updated the gloss load script.


arthurian (approved):
Looks good! :+1:

PR #440, opened 2022-10-21
fix personal vocab edit

Tanvez:
Remove comma in `personalvocabentries` model because the `lemma_id` was being transformed into a tuple.


Tanvez:
Merging this in as this is a small change that me and @jaguillette looked into.

PR #443, opened 2022-10-28
Fix familiarity in learner view

Tanvez:
This PR fixes the text learner view familiarity interaction. When changing a familiarity rating for new words that are not included in the users personal vocabulary list the UI would not update the ratings for the newly changed/added word when navigating away and back to the previously changed/added word.

Note: the updated code is very similar to how `UPDATE_VOCAB_ENTRY` actions is interacting with the `updatePersonalVocabList` api. We will need to optimize the code to conform to DRY principles

### DIFF
--- a/static/src/js/app/vuex/actions.js
+++ b/static/src/js/app/vuex/actions.js
@@ -46,11 +46,15 @@ export default {
       .fetchVocabLists(state.text.lang, (data) => commit(FETCH_VOCAB_LISTS, data.data))
       .catch(logoutOnError(commit));
   },
-  [CREATE_VOCAB_ENTRY]: ({ commit, state }, { lemmaId, familiarity, headword, definition }) => {
-    const cb = (data) => commit(FETCH_PERSONAL_VOCAB_LIST, data.data);
-    api
-      .updatePersonalVocabList(state.text.id, lemmaId, familiarity, headword, definition, null, null, cb)
-      .catch(logoutOnError(commit));
+  [CREATE_VOCAB_ENTRY]: async ({ commit, state }, { lemmaId, familiarity, headword, definition }) => {
+    // TODO: Make DRY with updateVocabEntry
+    const { response, data } = await api
+      .updatePersonalVocabList(state.text.id, lemmaId, familiarity, headword, definition, null, null);
+    if (response && response.status >= 400) {
+      return response;
+    }
+    commit(FETCH_PERSONAL_VOCAB_LIST, data);
+    return null;
   },
   // eslint-disable-next-line max-len
   [UPDATE_VOCAB_ENTRY]: async ({ commit, state }, { entryId, familiarity, headword, definition, lang = null, lemmaId }) => {


static/src/js/app/vuex/actions.js
```diff
@@ -46,11 +46,15 @@ export default {
...
   },
-  [CREATE_VOCAB_ENTRY]: ({ commit, state }, { lemmaId, familiarity, headword, definition }) => {
-    const cb = (data) => commit(FETCH_PERSONAL_VOCAB_LIST, data.data);
-    api
-      .updatePersonalVocabList(state.text.id, lemmaId, familiarity, headword, definition, null, null, cb)
-      .catch(logoutOnError(commit));
+  [CREATE_VOCAB_ENTRY]: async ({ commit, state }, { lemmaId, familiarity, headword, definition }) => {
+    // TODO: Make DRY with updateVocabEntry
+    const { response, data } = await api
+      .updatePersonalVocabList(state.text.id, lemmaId, familiarity, headword, definition, null, null);
```
arthurian, line 52: In the old version of this code, a callback was passed to `updatePersonalVocabList()` and the `commit()` happened in the callback, but in the revised code, it seems the callback has been removed and the commit is happening after a successful API response (<400). I'm just curious, but was that the issue, the callback wasn't being called and therefore the commit wasn't happening?

static/src/js/app/vuex/actions.js
```diff
@@ -46,11 +46,15 @@ export default {
       .fetchVocabLists(state.text.lang, (data) => commit(FETCH_VOCAB_LISTS, data.data))
       .catch(logoutOnError(commit));
   },
-  [CREATE_VOCAB_ENTRY]: ({ commit, state }, { lemmaId, familiarity, headword, definition }) => {
-    const cb = (data) => commit(FETCH_PERSONAL_VOCAB_LIST, data.data);
-    api
-      .updatePersonalVocabList(state.text.id, lemmaId, familiarity, headword, definition, null, null, cb)
-      .catch(logoutOnError(commit));
+  [CREATE_VOCAB_ENTRY]: async ({ commit, state }, { lemmaId, familiarity, headword, definition }) => {
+    // TODO: Make DRY with updateVocabEntry
```
arthurian, line 50: Good point -- definitely worth revisiting on a subsequent update.

arthurian (approved):
Looks good @Tanvez! :+1:

static/src/js/app/vuex/actions.js
Tanvez: The short version, the callback function was removed from the `updatePersonalVocabList` API with the addition of showing error messages in the UI and that cause this bug here since another action(`CREATE_VOCAB_ENTRY`) was using the same API call and the code was never updated with the removal of the cb function.
#### A little longer explaination
The change here is mainly due to the `updatePersonalVocabList` API being used in two places, `UPDATE_VOCAB_ENTRY` and `CREATE_VOCAB_ENTRY`. The initial change was made to the `UPDATE_VOCAB_ENTRY` and that included a removal of the Callback function due to the addition of the alert(`toast`) message for failing/successful edits in the `personal vocab list`. I thought it would be clearer to convert the function to use async/await than using a callback function(i think async/await was the solution to cbs) to handle passing along the `status codes` to the UI.

We can definitely split these API calls out into its own function if we want to preserve the original logic.

jaguillette (approved):
Looks good to me! The conversation on the fix was helpful, thanks!

static/src/js/app/vuex/actions.js
arthurian: That makes total sense, thanks! :+1:

PR #444, opened 2022-10-28
Vocab split script

jaguillette:
This just adds a script to split an input file based on a "chapter" field to output a set of vocab list spreadsheets for each chapter. Each vocab list has the vocab for that chapter and every chapter before it. The script should be invoked by 

```python split_vocab_spreadsheet_by_chapter.py path/to/spreadsheet.tsv```


arthurian (approved):
Looks good @jaguillette. :+1:

Tanvez (approved):
LGTM!

PR #445, opened 2022-10-31
Consolidate vocab lists

jaguillette:
This PR makes a number of changes to consolidate the personal vocab list detail view with the regular vocab list detail view. The goal is to use the same Vue component, regardless of what kind of vocab list is being displayed.

To that end, the following changes have been made:
- Django:
	- Add an API endpoint at `api/v1/vocab_lists/<int:pk>` to get details for a given vocab list
	- Add a POST method to the vocab list entries endpoint at `api/v1/vocab_lists/<int:pk>/entries/` to allow creating a new vocabulary list entry
	- Direct both the `PersonalVocabListDetailView` and the `VocabularyListDetailView` to use the same `vocab_list/detail.html` template to render vocab list details
	- Modify the `vocab_list/detail.html` template to only display relevant controls for the kind of vocab list displayed, and pass along enough information for the Vue template to render the `PersonalVocab.vue` template.
	- Make the data output by personal vocab list entries and vocab list entries consistent so they can both be consumed by the same front-end.
- Vue
	- Update constants, actions, mutations, and api calls to support create, update, and delete operations on individual vocab entries
		- Mutations on any  kind of vocab list in a detail view should update the `vocabList` variable in state.
	- Modify `PersonalVocab.vue` with conditional logic based on the kind of vocab list being displayed, based on the `personalVocab` variable passed to the div that creates the view.
		- Familiarity interface elements should only be displayed for personal vocab lists
	- Update `QuickAddVocabForm.vue` to work with regular vocab lists as well as personal vocab lists. This includes hiding familiarity rating when adding a new entry.
	- Update tests to pass with the new changes.

Note that after review, I'll move the contents of `PersonalVocab.vue` into `Vocab.vue`, update the necessary references, and remove `PersonalVocab.vue`. I've left the changes on `PersonalVocab.vue` since that is the template that is being applied to regular vocab lists, so it's easier to look at this way.


static/src/js/app/vuex/actions.js
```diff
@@ -39,13 +46,34 @@ export default {
...
     .fetchText(id, (data) => commit(FETCH_TEXT, data.data))
     .catch(logoutOnError(commit)),
-  [FETCH_VOCAB_LISTS]: ({ commit, state }) => {
-    api
-      .fetchVocabLists(state.text.lang, (data) => commit(FETCH_VOCAB_LISTS, data.data))
+  [FETCH_VOCAB_LIST]: async ({ commit }, { vocabListId }) => {
+    const { data } = await api
+      .fetchVocabList(vocabListId)
+      .catch(logoutOnError(commit));
+    commit(UPDATE_VOCAB_LIST, data.data);
```
Tanvez, line 53: Just thinking for future improvements we could alert the user if there is an error fetching the `vocab list`

static/src/js/app/vuex/actions.js
```diff
@@ -68,20 +96,63 @@ export default {
...
+     * it doesn't need to look up the new entry in the results.
+     */
+    const { entries: localEntries } = state.vocabList;
+    const localEntryIndex = localEntries.findIndex((el) => el.id === entryId);
+    // Work with the entries only, rather than the whole vocab list, since
+    // working with the whole list results in the table not showing the updated entry.
+    const updatedEntries = [...state.vocabList.entries];
+    updatedEntries[localEntryIndex] = data;
+
+    console.log('updatedEntries:');
```
Tanvez, line 152: We should remove any of the debugging console.logs

vocab_list/views.py
```diff
@@ -23,13 +23,23 @@ class VocabularyListDetailView(DetailView):
...
         )
         return queryset
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        vocab_list = self.get_object()
+        context.update({
+            "pagetitle": vocab_list.title
+        })
+        print(context)
```
Tanvez, line 40: Should remove any print statements

static/src/js/app/vuex/actions.js
jaguillette: Yeah, it seems like a different way of surfacing errors might be a good thing to set up, maybe in a v1.1

arthurian (approved):
@jaguillette Great work consolidating the general and person vocab UIs! I just made a few minor suggestions, but nothing blocking. The use of the `isPersonalVocab` boolean in the Vue component is effective to orchestrate actions and enable/disable different UI features, and I think that may motivate more consolidation down the line. The additional documentation and unit tests are also fantastic. :+1:

hedera/api.py
```diff
@@ -327,6 +395,9 @@ def get_object(self):
             lang = self.text.lang
         else:
             lang = self.request.GET.get("lang")
+            # Try to access lang as a supported language to throw a key error
+            # if it isn't there
+            SUPPORTED_LANGUAGES[lang]
```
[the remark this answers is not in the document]
jaguillette: Yeah, in testing a blank `lang` value was created on a vocab list, and it wrought havoc throughout the app, so I definitely don't want that to happen again!

PR #447, opened 2022-11-04
Fix failing python unit tests

arthurian:
This PR fixes some failing python unit tests and updates the email confirmation setup for new accounts.

Notes:
- The [django-user-accounts](https://github.com/pinax/django-user-accounts) module is used to manage user accounts, including registration. To require an email confirmation, we just need to enable [ACCOUNT_EMAIL_CONFIRMATION_REQUIRED](https://django-user-accounts.readthedocs.io/en/latest/settings.html#account-email-confirmation-required) in `hedera/settings.py`.
- We may want to be careful about using `get_object_or_404()` in API views, since that returns an HTML response, not a JSON response (this was causing one of the API tests to fail). 
- A number of vocab tests were failing due to a shared reference to a CSV file. The reason is that after the first `read()` of the file, all subsequent reads return no data unless the file position is reset to the beginning of the file (e.g. `seek(0)`). The fix is to either reset the file position in each test, or create a separate file object for each test (I chose the latter).
- Hedera includes a centralized logging configuration, so we can replace `print()` statements with calls to `logger.info()` or `logger.debug()` where needed.


jaguillette (approved):
Looks good to me! The patterns around error handling in the API and file loading in tests look good, glad to have a solid base to build on.

Tanvez (approved):
Looks good to me!

PR #448, opened 2022-11-04
Learner Selected Word Reveal Gloss

arthurian:
This PR updates the learner view of the selected word so that the dictionary entry is always displayed, and only the gloss is hidden by default.

**Default view when clicking on a word in the text:**

selected-word-default

**With the gloss revealed:**

selected-word-gloss-revealed


Tanvez (approved):
LGTM!

bilbe (approved):
Looks great!

PR #453, opened 2022-11-07
Update Latin Import Data from DEV

arthurian:
This PR refreshes the import data based on changes made in the DEV database. This is in preparation for deploying the MVP candidate to the PROD environment.


Tanvez (approved):
LGTM!

jaguillette (approved):
Looks good to me! Do you have a script that you used for the diff against the data you pulled from the db? If it's easy to add it would be nice to have in the repo.

arthurian:
@jaguillette Good point. I didn't use a script for this, so I just updated the description with the process I followed. It could be scripted if this is more than a one time thing.

bilbe (approved):

PR #454, opened 2022-11-08
Fix js test

Tanvez:
### This PR fixes the JS tests and added backed focus functionality to the quick add component.
- Fix test for `actions`, `quickAdd` and `personalVocab` components
- Add back focus when quickAdd button is clicked to re-enable esc/exit functionality
- Removed lattice from code
- Removed console logs
- Minor changes to quick add(`v-if`, `Object.prototype.hasOwnProperty.call`)
#### Notes
- there is a vue warning when running js test `[Vue warn]: Unknown custom element: <b-form-input> - did you register the component correctly? For recursive components, make sure to provide the "name" option.` This is mainly due to bootstrap-vue. Still need to look into why that warning is showing but it is not breaking.


static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
```diff
@@ -244,7 +243,8 @@
         }
 
         // Get headword from database if it isn't already in state
-        if (!Object.hasOwn(this.$store.state.forms, this.headword)) {
+        // Note: Node v14.X does not have a function called Object.hasOwn so converted to hasOwnProperty
```
arthurian, line 246: Interesting note about `Object.hasOwn`. According to https://v8.dev/features/object-has-own it's only available in V8 versions >= 9.3, which means [NodeJS 16.x or higher](https://nodejs.org/en/download/releases/).

arthurian (approved):
The updated JS tests look good to me, as well as the additional cleanup of the lattice stuff and removal of extra/unnecessary logging. 👍

static/src/js/app/components/quick-add-button/QuickAddVocabForm.vue
Tanvez: Yea, I only encountered this problem when I tried running the tests and the environment fail when accessing the quickAdd component since I'm locally running on `Node 14.x` but would otherwise pass if switched to a higher Node version. However when the code is ran in docker everything seems to work as expected with `Object.hasOwn`. If we ever upgrade the Node version in the future we can update accordingly.

jaguillette (approved):
Looks good to me! Good to have those lattice functions totally removed.

Issue #460, opened 2022-11-10
Everything getting marked unknown after disambiguation


IvyJL:
A few unknown words were displayed. Then I disambiguated one. And suddenly everything went unknown.
<img width="1148" alt="Everything unknown after disambiguation" src="https://user-images.githubusercontent.com/18600875/201169237-8f18db6b-a9a5-4c0d-8146-e03aac0c87d4.png">

IvyJL:
So a list is selected and "show unknown" is clicked. But when you disambiguate something, it's like the list gets unselected, the percents go back to 0% and everything is considered unknown. If you reselect the list, then you get updated percents. I thought it used to just update the percentage of the selected list when you disambiguated.

bilbe:
Yes, I noticed yesterday that selecting my personal list showed everything as 0% and red, when, if fact, I new all the words. Definitely needs fixing

bilbe:
This is still broken. The way to reproduce this is to select a vocab list, select a word, and then disambiguate the word. At that moment all the words turn red.

IvyJL:
Seems to be fixed! Thanks!

PR #461, opened 2022-11-10
fix key error in API when edit happens

Tanvez:
### This PR fixes the bug #460 which would reset known/unknown words when text is compared to a personal vocab list.

This bug occurs when text is edited. The new tokens don't have "resolved" `key:value` pair in the newly created dict. This would throw a key error.
- Added conditional to check if dict has "resolved" key and default value for "resolved" since it is used for comparison further down in the code.
#### Note:
The newly calculated percentage don't match because of another bug where edited text have extra blank tokens

### DIFF
--- a/hedera/api.py
+++ b/hedera/api.py
@@ -215,11 +215,15 @@ class LemmatizationAPI(APIView):
                 token.update(lemma.gloss_data())
             if vocablist_id is not None:
                 vocab_entry = vocablist.entries.filter(lemma_id=token["lemma_id"])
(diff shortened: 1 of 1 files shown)


jaguillette (approved):
Nice find! That's a pretty elegant fix for giving it a default, too.

bilbe (approved):
I know we don't have to work again until Monday, but is there any chance we could get this merged into dev and then into production today? I'll understand if the answer is no.

Issue #463, opened 2022-11-11
Teacher view percentages still not right


bilbe:
Teacher view percentages still not right. Proper nouns are not being included.

![image](https://user-images.githubusercontent.com/5712116/201425932-bf894871-f9c8-4328-9597-151584d1cb6c.png)

I know all of these words according to my personal vocabulary.

IvyJL:
I think this goes beyond proper nouns. In the attached images, I compare the same text  (no proper nouns) with two vocab lists (Shelm 1-2 and Shelm 1-3). Both show the same number of unknown words in red, but the percentages are different.
<img width="1146" alt="Screen Shot 2022-11-14 at 2 16 30 PM" src="https://user-images.githubusercontent.com/18600875/201746852-54391ef6-fe38-489a-817d-afbca4e44cb0.png">
<img width="1157" alt="Screen Shot 2022-11-14 at 2 16 48 PM" src="https://user-images.githubusercontent.com/18600875/201746858-c8176c49-6f33-4ab9-b9ff-96e3bf5d34f9.png">

bilbe:
Hey, @IvyJL, can you cut and paste a text version of this "fabula rosae"?

IvyJL:
femina rosam videt. rosa iacet. femina rosam amat. "rosam iuvare debeo." femina nautam videt. nauta laborat. "nauta aquam habere debet." femina nautam vocat. "habesne aquam?" nauta tacet. "rosa iacet. rosam amo. iuvare rosam opto. optasne iuvare?" "rosam iuvare opto. habeo aquam. aquam habeo. habesne pecuniam?" femina tacet. "habes pecuniam! pecuniam opto." nauta feminam terrere optat. "timere debes! moneo te!" non timet femina nautam. nautam femina monet: "te moneo. debes timere." agricola laborat. agricola feminam amat. feminam agricola videt. nautam agricola videt. agricola timet. feminam iuvare optat. vocat agricola feminam: "heus!" agricola feminam monet. "cave nautam!" agricola non timere debet. femina nautam superat! "supero." nauta iacet. nautam agricola videt. "fortuna te iuvat!" "habeo fortunam. habesne pecuniam?" "pecuniam habeo." femina aquam habet. "pecuniam debeo."

IvyJL:
@Ashley found that the discrepancy in the "fabula rosae" percentages was due to a lemmatization issue with the enclitic -ne. I did a quick test of a very short text and the percentages seem fine, so closing.

PR #467, opened 2022-11-14
fix disamb bug

Tanvez:
This PR fixes the bug found here #460 where the known/unknown words are reset after disambiguating a word when a vocab list is selected and unknown/known calculation to include proper nouns - issue #463.

Note: still investigating the different percentages Ivy showed in #463. Possible cause - If she edited the text the percentages will be different due to the bug with text edit


arthurian (approved):
Looks good @Tanvez 👍

jaguillette (approved):
Nice, looks good to me!

PR #472, opened 2022-11-23
fix API check for invalid language

jaguillette:
This is from Artie's feedback from a previous PR (#445) where I didn't notice I hadn't saved the file before committing


arthurian (approved):
LGTM! 👍

PR #474, opened 2022-11-30
fix extra tokens appearing after edits

Tanvez:
This PR fixes the extra tokens that are being generated when newlines are part of the text format(e.g. poems).
- Check have been added to `def lemmatize_chunk()` function to check if the chunk string equals `\r\n`(carriage return and newline) if that is true it will not add the tokens to the `lemmatized_text_data` list but instead add the `\r\n` to the token at the end of `lemmatized_text_data`
-  `def parse_following()` have also been updated to replace `\r\n` in addition to `\n`
- Test for the edit functionality is added(`test_handle_edited_data_no_changes`). This test will match token count to `test_lemmatized_text`
## Note: 
- I thought that making a change to lemmaize_chunk would lessen the impact of this particular change to other parts of the codebase and to keep the lemmatizer/tokenizer dumb.
- Left `initial` attribute in the lemmatized text data dict for future upgrades to include newlines/punctuations at the beginning of text
## TODO:
- Add test for parser


arthurian (commented):
@Tanvez this looks pretty good, but I did notice two issues that should be addressed before we merge:

1. An error is raised if the user inserts line breaks at the beginning of a text.
2. The `gloss_ids` are not being parsed correctly after the edits are completed (they're being saved as strings `"[]"` instead of lists `[]`), which is affecting the lemmatized text view. 

Lastly, I noticed that in `handle_edited_data()` we're replacing `<br/>` with `\n`, but not doing the same for `\r\n`. Would it be possible to do those transformations before handing it off to the `EditedTextHtmlParser`? You may have already considered that, but just wanted to check.

jaguillette:
I'm seeing the same issues that Artie mentioned about gloss lists being wrapped in quotes and newlines at the start of a text throwing errors. I don't see immediate fixes, but I'll take a closer look. Might be something to take a look at in a dojo?

lemmatized_text/parsers.py
```diff
@@ -18,13 +19,12 @@ def __init__(self, token_lemma_dict=None, lang=None):
     def handle_starttag(self, tag, attrs):
         if tag == "span":
             self.current_tag = "span"
-            self.current_attrs = dict(attrs)
+            self.current_attrs = json.loads(attrs["data-token"])
```
arthurian, line 22: Note that the HTML escaping will need to be reversed before parsing the JSON, otherwise I think this will return a `JSONDecodeError`. Since `html.escape()` was used above, you should be able to do an [html.unescape()](https://docs.python.org/3/library/html.html#module-html).

lemmatized_text/parsers.py
```diff
@@ -23,7 +23,7 @@ def handle_starttag(self, tag, attrs):
     def handle_endtag(self, tag):
         if "follower" in self.current_attrs:
             self.separate_true_followers(self.current_data)
-        elif self.current_data is not None:
+        elif self.current_data is not None and self.current_tag is not None:
```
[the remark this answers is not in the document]
Tanvez: I see that in some cases the `handle_endtag` will get an empty/`None` `self.current_tag` when there is a newline/break in the text. See below for some logging outputs where current_tag is None after the end of the text.
```
hedera-django      | self.current_tag is span current_attrs {'glossed': 'glossed-automatic', 'initial': '', 'lemma_id': 1299, 'resolved': 'no-ambiguity', 'gloss_ids': [83198], 'word_normalized': 'venit'}
hedera-django      | self.current_tag is None current_attrs {}
hedera-django      | self.current_tag is span current_attrs {'follower': 'true'}
hedera-django      | self.current_tag is None current_attrs {}
hedera-django      | self.current_tag is None current_attrs {}
```

lemmatized_text/tests.py
```diff
@@ -124,6 +129,30 @@ def test_handle_edited_data(self):
...
+    def test_handle_edited_data_no_changes(self):
+        example_text = LemmatizedText.objects.create(
+            title="Test title",
+            lang="lat",
+            original_text=test_original_text,
+            created_by=self.created_user1,
+            data=test_lemmatized_text
+        )
+        example_text.handle_edited_data("Test title", test_original_text_html)
+        self.assertEqual(example_text.token_count(), len(test_lemmatized_text))
```
Tanvez, line 141: Add test to check token/word matches original

lemmatized_text/test_data.py
```diff
@@ -0,0 +1,130 @@
...
...&quot;word_normalized&quot;: &quot;Italiam&quot;}">Ītaliam</span><span follower="true">, </span><span data-token="{&quot;glossed&quot;: &quot;glossed-automatic&quot;, &quot;lemma_id&quot;: 1585, &quot;resolved&quot;: &quot;resolved-automatic&quot;, &quot;gloss_ids&quot;: &quot;[72637, 85017]&quot;, &quot;word_normalized&quot;: &quot;fato&quot;}">fātō</span><span follower="true"> </span><span data-token="{&quot;glossed&quot;: &quot;glossed-automatic&quot;, &quot;lemma_id&quot;: 39952, &quot;resolved&quot;: &quot;no-ambiguity&quot;, &quot;gloss_ids&quot;: &quot;[79267]&quot;, &quot;word_normalized&quot;: &quot;profugus&quot;}">profugus</span><span follower="true">, </span><span data-token="{&quot;glossed&quot;: &quot;na&quot;, &quot;lemma_id&quot;: null, &quot;resolved&quot;: &quot;no-lemma&quot;, &quot;gloss_ids&quot;: &quot;[]&quot;, &quot;word_normalized&quot;: &quot;Laviniaque&quot;}">Lāvīniaque</span><span follower="true"> </span><span data-token="{&quot;glossed&quot;: &quot;glossed-automatic&quot;, &quot;lemma_id&quot;: 1299, &quot;resolved&quot;: &quot;no-ambiguity&quot;, &quot;gloss_ids&quot;: &quot;[83198]&quot;, &quot;word_normalized&quot;: &quot;venit&quot;}">vēnit</span>'
```
Tanvez, line 130: use `transform_data_to_html function`, make sure we have test for `transform_data_to_html` as well

arthurian (approved):
Thanks @Tanvez! We looked at this synchronously this morning. The changes look good overall, just a few minor tweaks to the tests for future maintainability. 👍

jaguillette (approved):
Looks good to me based on our synchronous review this morning!

PR #476, opened 2022-11-30
Make actions use async

jaguillette:
This PR makes any action that makes an api call into an `async` function that `await`s the response from the api, rather than relying on passing callbacks through to the api calls.

It also revamps the actions test suite to ensure that the API is being called with the expected parameters, mock out axios calls to the api to simulate a returned data object, and ensure that the mutation in the action is also called with the expected parameters. The first test in the suite has extra documentation to use as a reference implementation for adding additional action tests beyond the 5 that are currently implemented.

### DIFF
--- a/static/src/js/app/vuex/tests/actions.spec.js
+++ b/static/src/js/app/vuex/tests/actions.spec.js
@@ -1,4 +1,6 @@
+import axios from 'axios';
 import {
+  PROFILE_FETCH,
   PERSONAL_VOCAB_ENTRY_CREATE,
   PERSONAL_VOCAB_ENTRY_DELETE,
   PERSONAL_VOCAB_LIST_FETCH_LANG_LIST,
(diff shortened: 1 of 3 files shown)


static/src/js/app/vuex/tests/actions.spec.js
```diff
@@ -62,38 +104,51 @@ describe('Actions', () => {
         lang: 'lat',
         vocabularyListId: 1,
       };
-      //   const response = { data: { created: true } };
+      // payload sent to api function
+      const { lemmaId, vocabularyListId, ...apiPayload } = payload;
+      apiPayload.lemma_id = lemmaId;
+      apiPayload.vocabulary_list_id = vocabularyListId;
```
arthurian, line 110: I was a little puzzled by this at first, not because the test is unclear, but because I wasn't sure why we couldn't just use the same payload for both the action and API request. If I'm understanding correctly, it's because we need to rename the parameters from _camelCase_ to _underscore_ format since that's what the API endpoints expect. Not a big deal, but seems like unnecessary complexity we've introduced inadvertently. As we continue to address tech debt, we may want to just use camelCase on the frontend so that way we don't need to do this kind of transformation. The API endpoints could do what they want after they've received the parameters, of course.

static/src/js/app/vuex/tests/actions.spec.js
```diff
@@ -1,59 +1,101 @@
...
+
+      // expect the API to have been called with the appropriate URL
+      expect(axios.get).toHaveBeenCalledWith(`${BASE_URL}me/`);
+
+      // expect the correct mutation to have been called with the correct update data
+      expect(commit).toHaveBeenCalledWith(
+        PROFILE_FETCH,
+        profile.data.data,
+      );
+    });
```
arthurian, line 60: This is a nice model of a test -- very clear and easy to follow!

arthurian (approved):
Looks good overall @jaguillette. I didn't see any issues with the switch to `async/await` in the API. It's nice to have that cleaned up and following the same conventions, so we don't have to pass callbacks around. Thanks for adding the actions test suite and building a solid reference implementation that can be extended further. The tests that you added all look good and are easy to follow. In fact, I think the tests are already highlighting additional opportunities for improvement down the line (nothing actionable for this review). Great work! 👍

static/src/js/app/vuex/tests/actions.spec.js
jaguillette: That's exactly it, those two keys just need to be renamed to use underscores when sending data to the API. Our js linter prefers camelCase variable names when possible, but on the django end underscores are used more often. Definitely something to make more consistent in the future.

static/src/js/app/vuex/tests/actions.spec.js
```diff
@@ -1,59 +1,101 @@
...
+  function expects. Note that the body of a response is in the `data` key of an
+  axios response, so most API calls will make use of that key. Reference the
+  axios response schema for other parts of the response that may be used.
+
+  Ideally tests should be grouped by data models, similar to the way that the
+  actions themselves are grouped, and each `describe()` should include some
+  shared objects for expected payloads and responses. Unfortunately
+  idiosyncracies in how the Django API is constructed make this impractical in
+  many cases, but it's something to shoot for. */
+  describe('hedera.Profile', () => {
```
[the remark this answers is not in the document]
Tanvez: Ditto! Nice detailed overview of the vuex + API testing structure!

Tanvez (approved):
Nice improvements to the actions and test! @jaguillette

arthurian:
Note: Bill, Vesna, and I discussed this PR in Hedera dojo. Going to go ahead and merge.

PR #478, opened 2022-12-02
Fix failing python tests

jaguillette:
For some reason, the Python tests have started failing in GitHub actions. A failing run with logs can be seen here: https://github.com/Hedera-Lang-Learn/hedera/actions/runs/3603801082/jobs/6073350159. The issue seems to be with setting up `paddle`, which is being installed because of `lac`, which is part of the lemmatizer for Chinese language text. Since we've removed Chinese language support from the app for the time being, this PR removes `lac` from the requirements and comments out the tests that depend on it. The library and tests should be reinstated as part of bringing Chinese language support back in to the app, when that happens.

### DIFF
--- a/hedera/requirements/base.in
+++ b/hedera/requirements/base.in
@@ -17,7 +17,7 @@ sentry-sdk~=1.3.1
 rq~=1.9.0
 wagtail~=2.14.1
 django-lti-provider~=0.4.7
-lac~=2.1.2
+# lac~=2.1.2
 pip-tools~=6.4.0
 tqdm~=4.63.0
 # Pinned protobuf to 3.20.1 because of LAC issue with protobuf 4.21.1
--- a/hedera/requirements/base.txt
+++ b/hedera/requirements/base.txt
@@ -107,8 +107,6 @@ jmespath==0.10.0
     #   botocore
 l18n==2021.3
     # via wagtail
-lac==2.1.2
-    # via -r hedera/requirements/base.in
 nameparser==1.1.1
     # via django-lti-provider
 numpy==1.21.6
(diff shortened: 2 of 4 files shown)


jaguillette:
`[lac](https://pypi.org/project/LAC/)` hasn't had updates since May 2021, so it isn't a recent update to that library, but one of its dependencies that my local environment picked up, `[paddlepaddle](https://pypi.org/project/paddlepaddle/)` got an update on November 17 of this year, so maybe that did it somehow? `paddlepaddle` isn't in `requirements/base.txt` though.

elliottyates:
Makes sense to me. LAC was brought on expressly for a POC; it seemed best positioned when we were looking at it to just show that the functionality is possible, but if concerted effort was made to really get Chinese in a good, scalable and sustainable state, it would probably merit re-evaluating the library landscape.

LAC doesn't pin `paddle` in any way:
https://github.com/baidu/lac/blob/master/python/setup.py

It's not clear to me from scanning the `paddle` releases or the LAC issues whether it's fixable, or what's causing it, but LAC certainly doesn't look maintained at the moment, judging by the Issues activity; the only issues closed since early 2021 have been by the issue creators themselves. 

Some relevant issues from LAC:
* An example, from before `paddle 2.4.0` (the Nov 27 release), of the same issue we're seeing: https://github.com/baidu/lac/issues/256. 
* Feb 2021 issue asking about support for Python 3.8 and paddle 2.0; creator closed it and said, essentially, "seems like it's working": https://github.com/baidu/lac/issues/175
* https://github.com/baidu/lac/issues/256, opened within the past 24h, basically just says "anyone maintaining this anymore?"

arthurian (approved):
Thanks @jaguillette. Disabling the `lac` dependency and commenting out the chinese tests seems OK for now. As Elliott pointed out, that dependency will need to be re-evaluated anyway when we integrate it back in in along with the rest of the code that was dependent on the lattice structure.

PR #483, opened 2022-12-06
fix texts list ui

Tanvez:
This PR fixes the lemmatized text list in the `Text vue` so all the lemmatized text are being populated correctly.


arthurian (approved):
LGTM! :+1:

Issue #473, opened 2022-11-29
Student View: not remembering familiarity


bilbe:
So the Student view forgets a words familiarity as soon as you move away from it, but it does add it to your Personal Vocabulary. Plus if you go back and add the familiarity again, it add the word again to your Personal Vocabulary.

NB, this is working in PROD but broken in DEV.

On the up side, it is very smooth deleting the extra entry in the Personal Vocabulary view.

IvyJL:
I think this is fixed now in dev. @bilbe , can you confirm?

bilbe:
Yes. This is definitely fixed

PR #484, opened 2022-12-06
Dojo 20221206

arthurian:
This PR is to fix the actions/mutations that are not working.
- Fixed issue with Learner view familiarity not updating correctly/duplicating entries in personal vocabulary view #473 
- Fixed issue with selecting a new gloss in lemmatized text view causing the text to disappear
- Added `LEMMATIZED_TEXT_FETCH_LIST ` action and mutation for global use
- Fixed/rename actions and mutations related to personal vocabulary list
- Fix payload names so mutations receive the correctly structured payloads
- Fix failing mutations, actions and Personal Vocab tests

### DIFF
--- a/static/src/js/app/PersonalVocab.vue
+++ b/static/src/js/app/PersonalVocab.vue
@@ -380,6 +380,9 @@
       },
       vocabList() {
         // Retrieve vocab list from state
+        if (this.isPersonal) {
+          return this.$store.state.personalVocabList;
+        }
         return this.$store.state.vocabList;
       },
       vocabEntries() {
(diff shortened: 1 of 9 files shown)


arthurian (commented):
Looks good @Tanvez. Nice catch on some of the other issues, including the personal vocab lists. :+1:

jaguillette (approved):
Looks good to me! I had set the personal vocab and regular vocab actions and mutations to both act on the same `vocabList` in state, but I think that having separate entities in state and leaving it to components to treat the same works as well. It looks like you've made the data getting sent to mutations more consistent too, which seems good to me. Thanks!

Tanvez (approved):

PR #496, opened 2023-01-13
Fix bug where quick add would not load in dashboard and navbar

Tanvez:
This PR fixes a bug where quick add would not load in dashboard and navbar due to undefined variable `this.vocabListType` and failing submit for personal vocab list from quick add modal
 - Added secondary condition to default to personal vocab  if `this.vocabListType` is null
 - Fix constant name/action for `vocabularyListEntry_create`
 - Added `isPersonal` variables to dashboard and navbar to default to personal vocab list in the quick add modal
 - Fix quick add for personal vocab list


jaguillette (approved):
Looks good! Finally tested on my local install and was able to quick add vocab from the dashboard.

PR #503, opened 2023-02-06
Fix personal vocab lemma edit

### LINKED ISSUE #501: Editing lemma in personal vocabulary
It looks like you can edit the lemma, but changes do not take.

Tanvez:
This PR fixes #501  the personal vocab lemma edit by fixing the endpoint where we retrieve the lemma list for the autocomplete feature.


jaguillette (approved):
Reviewed in dojo, LGTM!

PR #508, opened 2023-02-15
Fix number of database queries from dojo

### LINKED ISSUE #507: Long load times and 503 errors in dev
When loading the dev instance, some page loads are taking longer than expected, and others are throwing a 503 error.

Tanvez:
This PR is from the dojo between me and @jaguillette(thanks to Jeremy for cracking the code/django docs on this) which reduces the number of database queries that `PartialMatchFormLookupAPI` makes. Fixes #507 
- The endpoint has been refactored to `filter`/`distinct`/`order_by`(rank) on the `Lemma` model based on the `forms__form` parameter via the  reverse relation from the `FormToLemma` model through the `related_name` attribute.
- `prefetch_related` has been implemented to retrieve the `glosses` associated with each of the lemma which drastically reduces the number of queries from over `9000` to `1`

### DIFF
--- a/hedera/api.py
+++ b/hedera/api.py
@@ -549,23 +549,14 @@ def get_any_vocablist_by_id(user_id, lang, vocablist_id):
 
 
 class PartialMatchFormLookupAPI(APIView):
+
     def get_data(self):
         form = self.kwargs.get("form")
         lang = self.kwargs.get("lang")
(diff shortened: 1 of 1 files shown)


arthurian (approved):
Looks good @Tanvez (and @jaguillette)! Nice work fixing the performance issue! :+1:

jaguillette (commented):
LGTM!

Issue #500, opened 2023-01-13
Adding underscore when editing text does not join


IvyJL:
After adding a text, I realized I had missed a two-part verb form (contristatus est). I went to edit and added the underscore (contristatus_est), but the text does not relemmatize.

In other words, I had contristatus est which lemmatized to contristo and sum.
I added the underscore so it would lemmatize to just contristo, but it is still lemmatized as contristo and sum.

I'm not actually sure if this was meant to be fixed now.

bilbe:
This pretty much works, BUT if you type in this:
`Carthāgō dēlenda est es. Caesar sum, esse, fui, futurus, est`

and then edit it to this:
`Carthāgō dēlenda_est es. Caesar sum, esse, fui, futurus, est`

You will only see `Carthāgō dēlenda est es...` But when you arrow key from `Carthāgō` to `dēlenda est`, you will see a disambiguation for `dēlenda`, `est`, and then `dēlenda est`. Try it and you'll see what I mean.

PR #509, opened 2023-02-24
fix underscore

Tanvez:
This PR is to fix issue #500 where latin text edited with underscores will not be picked up by the editor.
- added `check_text` and `apply_text_rule ` to latin service to be used in the `parsers.py` to determine underscores in a text
- added test for edits with underscore

### DIFF
--- a/lemmatization/services/base.py
+++ b/lemmatization/services/base.py
@@ -49,11 +49,19 @@ class BaseService(object):
 
     def check_text(self, text):
         """
-        Returns text or False of text
+        Returns TRUE or False based on the
 
         This method must be overridden.
         """
         raise NotImplementedError("Must provide check_text implementation on a per service basis.")
+
+    def apply_text_rule(self, unique_text, data):
+        """
+        Returns text or dict based on rules applied to modify the text
+
+        This method must be overridden.
+        """
+        raise NotImplementedError("Must provide apply_text_rule implementation on a per service basis.")
 
 
 class HttpService(BaseService):
--- a/lemmatization/services/base.py
+++ b/lemmatization/services/base.py
@@ -49,9 +49,8 @@ class BaseService(object):
 
     def check_text(self, text):
         """
-        Returns TRUE or False based on the language rules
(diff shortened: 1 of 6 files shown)


lemmatization/services/latin.py
```diff
@@ -96,6 +96,22 @@ def normalize(self, word):
...
 
+    def check_text(self, text):
+        """
+        For Latin, checking if edited text needs to be formatted based on underscore via latin_periphrastic_normalizer
+        """
+        if "_" in text:
+            return text
+        return False
+
+    def apply_text_rule(self, unique_text, data):
```
arthurian, line 107: Add a docstring to this function if possible (and add to base class).

lemmatization/services/base.py
```diff
@@ -47,6 +47,22 @@ def normalize(self, word):
         """
         return word
 
+    def check_text(self, text):
+        """
+        Returns TRUE or False based on the
```
jaguillette, line 52: "...rules applied to modify the text", like the next function?

Tanvez: Good catch! I was referring to the rule for the text based on specific text(e.g. latin uses and underscore to join specific lemmas together). Not necessarily modifying it for this function but we could combine the two functions in the future.
I'm mainly using this function to see if we should keep a record of the `chunk` to be joined later in the code

jaguillette (approved):
Just noticed the one other typo, otherwise looks pretty good!

lemmatized_text/parsers.py
```diff
@@ -49,6 +53,11 @@ def handle_endtag(self, tag):
         self.current_data = ""
 
     def handle_data(self, data):
+        # used to modify data by the service(e.g latin underscores)
+        formated_text_data = self.service.apply_text_rule(self.unique_text, data)
```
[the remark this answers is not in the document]
Tanvez: Nice catch!

arthurian (approved):
Overall looks good @Tanvez! 👍

Issue #499, opened 2023-01-13
Dev lemmatization progress bar broken


IvyJL:
Lost this on dev somehow. I have to reload the Texts page to see when lemmatization is done.

bilbe:
This is still broken!

bilbe:
Here's a text that you can use to reproduce the problem

Arma virumque canō, Trōiae quī prīmus ab ōrīs
Ītaliam, fātō profugus, Lāvīnia|que vēnit
lītora, multum ille et terrīs iactātus et altō
vī superum saevae memorem Iūnōnis ob īram;
multa quoque et bellō passus, dum conderet urbem,
inferretque deōs Latiō, genus unde Latīnum,
Albānī|que patrēs, atque altae moenia Rōmae.

bilbe:
Not really fixed. 

Well, it better and it does finish, which is critical. However, I tried some longer texts and it seem not to truly "progress." It seems that it just has one position initially and then complete. It does the job but is kind of unsatisfying.

arthurian:
The UX should be improved now with #521, which animates the progress bar between status updates so it doesn't just jump from like 27% complete to 100% complete (especially for shorter texts).

bilbe:
This works now

PR #511, opened 2023-03-02
fix progress bar for text lemmatizer

Tanvez:
This PR fixes the progress bar indicator when submitting text to be lemmatized.
- Changed the names of the API functions to match up to the newly formatted names


arthurian (approved):
LGTM :+1:

jaguillette:
This looks like a fix was implemented for #499

PR #513, opened 2023-03-08
fix failing editor when unedited text are submitted

Tanvez:
This PR fixes the failing edits when unedited text is submitted to the edit endpoint.
- Added a check for length of the new_data due to error `list index out of range` when accessing array elements in `parsers.py`


arthurian (approved):
LGTM 👍

PR #518, opened 2023-03-10
Fix lemma edit

Tanvez:
This PR fixes the lemma edit functionality so that lemmas with numbers attached can be use in the personal vocab editor.


arthurian (approved):
Looks good -- great work @Tanvez! 👍

jaguillette (approved):

PR #521, opened 2023-03-14
Animate lemmatization progress bar

arthurian:
This PR adds a transition between progress bar status updates so that it looks smooth instead of jumping between completion states.

Thanks @Tanvez @jaguillette @bilbe (dojo style)!


Tanvez (approved):
LGTM!

PR #522, opened 2023-03-15
Add backup latin data 3-15-2023

Tanvez:
This PR adds backups of `form_to_lemma`, `lemma`, and `glosses` to the latin data folder.


jaguillette (commented):
The data looks good to me! Would it be worthwhile to add a README to this folder with some notes on what's here and a reference for how to pull this data again later?

Tanvez:
> The data looks good to me! Would it be worthwhile to add a README to this folder with some notes on what's here and a reference for how to pull this data again later?

Good call!

jaguillette (approved):
Just checked the k459 wiki notes, and they look good! Thanks!

arthurian (approved):
LGTM - just wondering if we still need this in light of #523? I'd probably lean toward storing our own instance backups separately unless there's a requirement to store it in the repo itself.

Tanvez:
@arthurian It does seem that we are repeating work here. I will close this since we already have another PR which address the three tables(form_to_lemma, lemma, and glosses)

PR #523, opened 2023-03-22
Tidy latin data load scripts

Tanvez:
This PR is to update the load scripts with newly updated latin data that was exported from Production database.
- removed unused files in `/data/lat`
- updated load scripts to use new files
- added a note to k459 wiki about how to update files via exports from sql queries


jaguillette (approved):
Looks good to me! Thanks for the wiki documentation too!

arthurian (commented):
Overall the changes look good! Just a few comments/questions:

1. **Including DB primary keys?**. I noticed that `latin_lemma_table.tsv` includes IDs from the database, but `latin_gloss_table.tsv` does not. We should probably be consistent here in how the data is represented, and whether we are intending to include DB artifacts (primary keys). In the case of the `latin_lemma_table.tsv`, the `lemma` field should be unique for a particular `lang`, so that's already a natural key. The load script doesn't seem to use either of those (id, lang), so they could probably be safely omitted and just rely on the natural keys.
2. **File naming conventions**. I don't think we gain a lot by prefixing the TSV files with `latin_*.tsv` since we know that all files in the `data/lat` folder are latin-specific already. Ideally, we should be able to have a standard convention such as:

    ```
    data/{lang}/lemma_table.tsv
    data/{lang}/gloss_table.tsv
    data/{lang}/form_to_lemma.tsv
    ```

    That would allow us to easily extend load scripts to other languages if they all follow that standard pattern. That's not to say there might not be other data files in those directories, but at a minimum, that standardization would be helpful.

Tanvez:
Good points @arthurian !
Definitely agree with the `DB primary keys` and naming conventions of the files.

arthurian:
Thanks @Tanvez! The changes look good 👍

jaguillette (approved):
Looks good to me, too!

Issue #525, opened 2023-04-19
Searching and sorting in Django interface


IvyJL:
It would be really helpful to be able to search and sort more tables so that corrections on the back end can be made more easily.

Vocabulary list entries: search and sort
Personalized vocabulary list entries: search and sort
Form to lemma: sort
Lemmatized texts: sort

An example of the issue: if I upload a new vocabulary list and need to make corrections to a lemma, I have to scroll through 50+ pages of Vocabulary list entries that are not in any particular order to find the word. Entries from the same list seem to get randomly distributed (see image).
<img width="1034" alt="Screenshot 2023-04-19 at 1 15 06 PM" src="https://user-images.githubusercontent.com/18600875/233167435-26d66e32-dcf3-49f5-8ff2-4fd8bd47ad25.png">

Tanvez:
@IvyJL I think you can click on the column titles(any table should have them) e.g. `FORM`, `LEMMA` and the rows will sort in descending or ascending order. Let me know if that fits your use case. Will look into adding the search option for the Vocabulary list entries  and Personalized vocabulary list
<img width="1483" alt="Screen Shot 2023-04-21 at 4 54 25 PM" src="https://user-images.githubusercontent.com/29421667/233733153-27e9b83f-4090-4346-9654-02e098d0102d.png">
 entries

IvyJL:
That does seem to work! I was sure it didn't before, but maybe it was just because the sorting arrows weren't visible.

PR #526, opened 2023-04-24
add searching for headwords in PersonalVocabularyList and Vocabulary …

Tanvez:
This PR adds `headword` search for `PersonalVocabularyList` and `VocabularyList` via the admin interface #525


jaguillette (approved):
Looks good to me! Double-checked locally, and it works great!

PR #528, opened 2023-05-10
Add sentry

Tanvez:
This PR adds sentry to frontend and backend.
- atg-ops-ecs has been updated to pull in sentry-auth-token
- added an endpoint to trigger via sentry-debug
- added sentry-auth-token into the ssm param store


jaguillette (approved):
LGTM!

PR #530, opened 2023-05-12
fixing undefined lang when loading vue dashboard

Tanvez:
This PR fixes an error where lang is undefined when `Dashboard.vue` is loaded.


jaguillette (approved):
LGTM!

PR #536, opened 2024-02-26
added default values for started_read_at and ended_read_at

Tanvez:
This PR is to fix the failing migration for the `started_read_at` and `ended_read_at`
- This error occurs when django applies the new migrations for the fields pictured above
`django.db.utils.IntegrityError: column "started_read_at" of relation "lemmatized_text_lemmatizedtextbookmark" contains null values`


jaguillette (approved):
LGTM!

PR #537, opened 2024-02-28
Updated migration files because of failing migration

Tanvez:
This PR is to update the migrations files due to the new `started_read_at` columns which will be created with null values. `null=True` was not part of the initial model creation/migration file which cause a postgres migration error.
- updated migration files
- removed unused migration files
Note: Deployed and tested on [dev.hedera](https://dev.hederaproject.org/)
Screenshot 2024-02-28 at 5 14 20 PM


arthurian (approved):
LGTM (reviewed synchronously) :+1:

